PR #8, opened 2023-02-03
Update Readme with benchmark output metric

johnugeorge:
Ref: #7 

Updating readme to reflect benchmark output metric


theoanab (approved):
thanks, Johnu!

PR #14, opened 2023-04-04
Adding report generation with sample results

johnugeorge:
The current PR adds two commands to benchmark script
1. datasize - for calculating minimum dataset size
2. reportgen - for generating report for single host and multi host run

Prefetch factor field is removed from open category submissions


theoanab (approved):
Everything looks good to me. I'm not sure that we need to have the sample results in the preview package.

PR #27, opened 2023-05-26
adding hack to add git hashing to report

jaredeh:
In the Storage WG meeting I made a suggestion for having the benchmark check the code.  This is an example of how that could work.  It will add a 'git' section to the .json report with a git hash and if relevant a diff showing what has been changed.


jaredeh:
recheck

PR #31, opened 2023-06-09
Update unet3d sample size

johnugeorge:
unet3d sample size is updated to match the reference benchmark


theoanab (approved):
We need to add this change to the first release, even if it is late, because the old sample size and stdev were not accurate. The sleep time of 1.36s corresponds to ~146MB sample size, with a batch size of 4, on V-100 GPUs.

PR #71, opened 2024-07-10
Update Submission_guidelines.md for V1.0 content.

FileSystemGuy:
This updates the file from v0.5 contents to be the v1.0 Rules.


FileSystemGuy (commented):
The updated dates are correct.

PR #84, opened 2025-02-19
feat: Add new playbooks for Ansible automation

glimchb:
- Introduced initial set of Ansible playbooks to automate the MLPerf storage benchmark tool.
- Playbooks include tasks for environment setup only for now.
- Playbooks install bare metal environment as well as docker engine.
- Added documentation for usage and configuration of the playbooks.

This commit enhances our automation capabilities, streamlining the benchmarking process and ensuring consistent, reproducible results.

PR #85, opened 2025-02-28
Benchmark.py script for v2.0

wvaske:
Benchmark script is getting migrated from bash to python for better integration with results checking scripts. 
- Update to latest version of DLIO
- Started updating rules document
- Separate config locations for training / checkpoint / vectordb


glimchb:
@wvaske 

since dlio now publishes to https://pypi.org/project/dlio-benchmark/2.0.0/
we can completely remove the submodule from here. it is not needed anymore...

and instead of 
```
pip3 install -r dlio_benchmark/requirements.txt
```
just do
```
pip3 install dlio-benchmark==2.0.0
```

wdyt ? much cleaner and more pythonic...

FileSystemGuy (approved):

wvaske:
@zhenghh04 , is the version of DLIO on pypi up-to-date with your changes? If not, can you rev the version to 2.1 or 2.0.1 and push a new version?

FileSystemGuy (approved):
Have you considered encoding some of the CLI options the user gives for "datagen" into a config file such that those option values are then inherited (and not changeable) when they start doing runs?  I'm a bit concerned that users will run datagen with some options and then run the benchmark with others (by accident).  Certainly the number of accelerators needs to be variable, but total DRAM size?

Are we going to (collect and/or) validate the DRAM size given by asking the benchmark clients at runtime their DRAM size?

In any case, I'm approving the change...

glimchb:
@FileSystemGuy @wvaske who can merge my #84 ? i can't even comment on that.... sorry for polluting your pr...

FileSystemGuy (approved):
Ah, I see you separated the SUT_dir from the results_dir (with better names) and are also on the path to record the operating config from one stage to the next.  That's what I get for reviewing them in time order rather than all at once.

Nit: the help text(s) includes "and data rates for each accelerator", which the user might confuse with the output number, which is a "data rate per accelerator".

Why does the datagen step care about the accelerator type?  If it generates a different dataset based upon the accelerator type, then the user should not be allowed to "run" with a different accelerator type than the dataset was generated for.  I didn't think datagen cared about accelerator type, so this may be moot.  Eg: is it kosher to to do a A100 "run" against against the same datagen'ed set of files as an H100  "run"?

FileSystemGuy (approved):

FileSystemGuy (approved):

FileSystemGuy (approved):

FileSystemGuy (approved):

FileSystemGuy (approved):
Ah, again I should have looked at everything before commenting...

FileSystemGuy (approved):
The attributes "per_host_mem_kB" and "total_mem_kB" are in kilo-bytes but the CLI args are in GB and the raw memory capacity pulled from the nodes info is in B.  Would standardizing all memory capacity variables on GB (or MB) be less risky for confusion?

FileSystemGuy (approved):

FileSystemGuy (dismissed):

FileSystemGuy:
Boris,

I don't see any problems from a "product owner" type of perspective, your change is "approved" in that sense, but I don't know if there are any technical comments that should be addressed before we merge.  That's up to Johnu and Huihuo and Wes.

Thanks,

Curtis
________________________________

FileSystemGuy (approved):

wvaske:
> The attributes "per_host_mem_kB" and "total_mem_kB" are in kilo-bytes but the CLI args are in GB and the raw memory capacity pulled from the nodes info is in B. Would standardizing all memory capacity variables on GB (or MB) be less risky for confusion?

My preference is to store information in it's "natural" form with a unit on the variable name so we can do math when we need to. But I'm open to changing.

FileSystemGuy (dismissed):

wvaske:
> As this PR appears to be updating the rules for v2.0, there was a recent discussion in the checkpointing subgroup about model sizes. The table below shows the memory requirements per node when using 32 nodes, displaying that the 1T parameter model far exceeds our available hardware (above 512GB/node). We would not be able to run the 1T model, so we recommend excluding it from the v2.0 rules **unless the calculations in this table are incorrect** (model size/ #node)
> 
> While the 1T model may be of interest to customers, it is impractical for us to run. For this reason, we would prefer not to include the 1T parameter model in the v2.0 submission guidelines (checkpointing).
> 
> The 405B model might already be at the limit of what we plan to run, as it would likely require 40 nodes instead of 32.
> 
> ```
> | Model          | SIZE (GB) | # Nodes | Memory/node (GB) |
> |----------------|-----------|---------|-------------------|
> | 8B (Zero3)     | 88        | 1       | 88                |
> | 70B            | 1100      | 32      | 34,375            |
> | 405B           | 6000      | 32      | 187,5             |
> | 1T             | 17000     | 32      | 531,25            |
> ```


In the previous release Huawei ran with 51 nodes and Nutanix ran with up to 66 nodes. So it's not unreasonable to expect that some submitters will have the systems available to run at the 1T level. 

And with the announcement of Llama 4 Behemoth clocking in at 2T parameters, the 1T might be 'small' for the benchmark.

Each of the sizes is optional. If a submitter wants to only run on 70b, that is perfectly acceptable. The 1T param model is difficult to represent.

FileSystemGuy (approved):

requirements.txt
```diff
@@ -0,0 +1,3 @@
+dlio-benchmark==2.0.0
```
glimchb, line 1: this is duplication if you already defined those in `pyproject.toml` above

pyproject.toml
```diff
@@ -0,0 +1,28 @@
...
+version = "2.0.0b1"
+description = "MLPerf Storage Benchmark Suite"
+readme = "README.md"
+authors = [
+    {name = "MLCommons Storage Working Group"}
+]
+
+dependencies = [
+#    "dlio-benchmark==2.0.0",
+    "psutil>=5.9",
```
glimchb, line 16: missing `pyarrow`

requirements.txt
```diff
@@ -0,0 +1,3 @@
...
+vectordb-bench==0.0.22
+psutil>=5.9
```
glimchb, line 3: missing `pyarrow`

mlpstorage/benchmarks/dlio.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def add_workflow_to_cmd(self, cmd) -> str:
+        # Configure the workflow depending on command
+        if self.args.command == "datagen":
+            cmd += " ++workload.workflow.generate_data=True ++workload.workflow.train=False"
+        elif self.args.command == "run_benchmark":
+            cmd += " ++workload.workflow.generate_data=False ++workload.workflow.train=True"
+
+        # Training doesn't do checkpoints
+        cmd += " ++workload.workflow.checkpoint=False"
```
ekaynar, line 138: needs to return cmd variable

pyproject.toml
```diff
@@ -0,0 +1,28 @@
...
+[project]
+name = "mlpstorage"
+version = "2.0.0b1"
+description = "MLPerf Storage Benchmark Suite"
+readme = "README.md"
+authors = [
+    {name = "MLCommons Storage Working Group"}
+]
+
+dependencies = [
```
glimchb, line 14: un comment dlio ?

mlpstorage/benchmarks/dlio.py
wvaske: Sometimes you find bugs and you wonder how it ever worked... I think this got caught in my refactor. Fixed in the latest commit

requirements.txt
wvaske: Added, thanks.

requirements.txt
wvaske: I use requirements.txt and pyproject.toml because pycharm doesn't know how to properly ready my pyproject.toml file. We can probably remove requirements unless we want to optimize the docker container build with a prereq install step

README.md
```diff
@@ -83,112 +83,290 @@ The working directory structure is as follows
```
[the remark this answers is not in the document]
wvaske: I went through and updated most of the README. Still need to finish up the individual workload instructions but it should be pretty clear how to use the program

wvaske:
Checkpointing works.
Check out the history function "mlpstorage history show"
Added a report generator. "mlpstorage reports reportgen"

Please test and provide feedback. The report generated has a lot of extra information that helps to know how the test was run W.R.T. input files, args, params, and how they combine. 

I recommend pulling the csv into Excel via power query so it's a data connection to a table. Then create a pivot table from the data table for doing analysis.

I currently capture cpu and memory information with passwordless ssh. If that doesn't work for you, please let me know so I can update with a different methodology.

Submission_guidelines.md
```diff
@@ -103,6 +103,7 @@ The following definitions are used throughout this document:
 - A **sample** is the unit of data on which training is run, e.g., an image, or a sentence.
 - A **step** is defined to be the first batch of data loaded into the (emulated) accelerator.
 - **Accelerator Utilization (AU)** is defined as the percentage of time taken by the simulated accelerators, relative to the total benchmark running time. Higher is better.
+- **Design power** is defined to be the minimum measurement of electrical power that must be capable of being supplied to a single or collection of power supply units (PSUs) in order to avoid violating regulatory and safety requirements. For individual PSUs, the design power equals the nameplate rated power. For groups of redundant PSUs, the design power is equal to the sum of the nameplate rated power of the minimum number of PSUs required to be simultaneously operational.
```
glimchb, line 106: 2.2 `Checkpointing` and 2.3 `Vector Database` above are empty. maybe at least add a TODO to TBD
maybe mention will be part of 3.0 release...

wvaske: Thanks. We'll remove vectorDB and Huihuo needs to add his proposed checkpoint rules document

pyproject.toml
```diff
@@ -0,0 +1,28 @@
...
+    "pyarrow"
+]
+
+[project.urls]
+"Homepage" = "https://github.com/mlcommons/storage"
+"Bug Tracker" = "https://github.com/mlcommons/storage/issues"
+
+[tool.setuptools]
+packages = {find = {}}
+
```
garimajain05, line 26: I pulled the changes locally and tried running the benchmark, but encountered the following error related to missing YAML configuration files:
```
FileNotFoundError: Configuration file not found: /root/checkpoint/mlperf-storage/.venv/lib/python3.12/site-packages/configs/dlio/workload/unet3d_h100.yaml
```
It looks like the YAML files under _configs/dlio/workload/_ are not being included in the installed package. As a result, they're not accessible at runtime.

To fix this, you might need to update **pyproject.toml** under **[tool.setuptools.package-data]** like so:
```
[tool.setuptools.package-data]
"mlpstorage" = ["../configs/dlio/workload/*.yaml"]
```
After making this change and reinstalling with **pip install .**, the benchmark worked as expected on my end

Submission_guidelines.md
zhenghh04: Yes, will add the checkpointing section soon.

PR #112, opened 2025-05-30
Wvaske/checkpoint fixes

wvaske:
To Do:
 - [ ] Add Rules Checker items for checkpointing
 - [ ] Verify Memory Capacity and that data written is bigger than memory or separate read and writes with call to clear caches in between and command to execute the writes gets printed
 - [ ] Check if data exists in checkpoint directory and raise & exit if so (require empty location)
 - [ ] Update rules document regarding above
 - [ ] Add info to rules that DLIO is the requirement and mlpstorage is the convenience methodology

Testing needed:
 - [ ] Verify subset checkpointing. mode=subset is automatically set when number of processes is less than required process count

### DIFF
--- a/configs/dlio/workload/cosmoflow_datagen.yaml
+++ b/configs/dlio/workload/cosmoflow_datagen.yaml
@@ -3,6 +3,11 @@ model:
 
 framework: tensorflow
 
+workflow:
+  generate_data: True
+  train: False
+  checkpoint: False
+
 dataset:
  data_folder: data/cosmoflow
  num_files_train: 524288
--- a/configs/dlio/workload/llama3_1t.yaml
+++ b/configs/dlio/workload/llama3_1t.yaml
@@ -1,6 +1,6 @@
 # we mimic the checkpoint data for megatron-deepspeed
 model: 
-  name: llama_405b
+  name: llama_1t
   type: transformer
   num_layers: 128
   model_datatype: fp16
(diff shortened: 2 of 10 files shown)


johnugeorge:
1. checkpoint folder for unet datasize command should be optional
2. datasize command output do not match with `mlpstorage reports reportgen` result
3. Checkpoining command run fails `mlpstorage checkpointing run --hosts 127.0.0.1 --model llama3-8b --num-processes 8 --num-checkpoints-read -1 --results-dir ./checkpoint_results_write --client-host-memory-in-gb 64 --checkpoint-folder checkpoints --num-checkpoints-write 2`

Can you also update Readme with checkpointing commands(like the training workloads) ?

johnugeorge:
datasize of unet3d fails for checkpoint-folder arf.  Can you add it to the test run script?

wvaske:
> 1. checkpoint folder for unet datasize command should be optional
> 2. datasize command output do not match with `mlpstorage reports reportgen` result
> 3. Checkpoining command run fails `mlpstorage checkpointing run --hosts 127.0.0.1 --model llama3-8b --num-processes 8 --num-checkpoints-read -1 --results-dir ./checkpoint_results_write --client-host-memory-in-gb 64 --checkpoint-folder checkpoints --num-checkpoints-write 2`
> 
> Can you also update Readme with checkpointing commands(like the training workloads) ?

For the checkpoint command, did you have "-1" for --num-checkpoints-read? If I use "1" instead of "-1" the command runs for me. 

And for the datasize output not matching, can you provide your specific example? I assume I'm reading the memory & host info from the DLIO logs incorrectly leading to a mismatch in the datasize output?

PR #163, opened 2025-06-17
System Configuration yaml and rules + bonus rule

wvaske:
Added clear rule on consecutive runs having less time between runs than there is time in a run
Updated system-configuration.yaml example file
Updated rule that system-configuration.pdf needs topology


TheKanter:
@wvaske this is awesome - thanks! Do we want to indicate that Rack Units are mandatory only for certain types of systems?

wvaske:
I merged main into the branch. It looks good to me and I think I addressed the issues here.

PR #167, opened 2025-06-18
system-configuration changes & mpi-options

wvaske:
Please test the --mpi-params option. I'm not scaled out so I'm not certain if it works in non-local use-cases properly. It should, it's just passing through from the mlpstorage cli to the mpirun cli.

Added cli option for --mpi-params. It can be used in the following ways:
```python
mlpstorage --mpi-params "--param1=value1 --param2=value2"
mlpstorage --mpi-params "--param1=value1" "--param2=value2"
mlpstorage --mpi-params "--param1-value1" --mpi-params "--param2=value2"
```

If the parameter doesn't begin with a hyphen, you don't need the quotes.

To see if it will work correctly use the `--what-if` flag to print the command that would be executed wtihout actually executing.

I still can't see @FileSystemGuy 's comments on the last issue. There was a 4th item I had to add to the system description that I didn't capture. Here's what I have:

```
System:
  name: FastAmazingAcmeStorage 9000
  description: <text>
  storage_location: [ remote | local | hyper-converged ]
  client_software: [ in-box | proprietary ]
  storage_interface: [ block | file | object ]
  required_rack_units:
  shared_capabilities:
    multi_host_support: True            # False is used for local storage
    ...
```

Let me know what I need to add.

### DIFF
--- a/mlpstorage/benchmarks/dlio.py
+++ b/mlpstorage/benchmarks/dlio.py
@@ -112,7 +112,8 @@ class DLIOBenchmark(Benchmark, abc.ABC):
         if self.args.exec_type == EXEC_TYPE.MPI:
             self.logger.debug(f'Generating MPI Command with binary "{self.args.mpi_bin}"')
             mpi_prefix = generate_mpi_prefix_cmd(self.args.mpi_bin, self.args.hosts, self.args.num_processes,
-                                                 self.args.oversubscribe, self.args.allow_run_as_root, self.logger)
+                                                 self.args.oversubscribe, self.args.allow_run_as_root,
+                                                 self.args.mpi_params, self.logger)
             cmd = f"{mpi_prefix} {cmd}"
 
         return cmd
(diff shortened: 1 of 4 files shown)


FileSystemGuy:
Here's the comments that didn't quite make it into PR #163:
 
Wes, everything else looks great, but I'm not sure about the types breakdown you have. Eg: software-defined is orthogonal to all of local, remote, and hyper-converged, it just means the vendor didn't sell the h/w the solution s running on. Parallel-FS is a very specific niche case, and shared-object is (nearly) a tautology, the point of object storage is to be massively shared without the locking overheads of POSIX.

Could we consider something closer to what we had before: local-block, remote-block, standards defined remote file (eg: NFS), proprietary remote file, object, other? My argument is that what the end-user cares about is the abstraction type (block, file, object) and the costs (local or remote, included in Linux or needs a proprietary plugin, needs dedicated H/W or can use shared H/W, etc).

We could treat it as a set of orthogonal pull-down menu choices:
- Local or remote
- Block, file, or object
- Included or not included in Linux
- Dedicated H/W or shared H/W
The above would represent all combinations. It would be displayed as "Local, block, included, dedicated H/W" or "Remote, file, not included, shared H/W", etc...

I'll note that we need to change the Section 3 definition of "storage system access type" in the Submission_guidelines.md file to match whatever we come up with here.

FileSystemGuy (approved):
Get the important changes into the script, and we may (or may not) come back to the SystemDescription type breakdown.

PR #183, opened 2025-06-30
Add a rules exception that it's OK to apply DLIO PR #299 in CLOSED

FileSystemGuy:
The following text change was applied: "The sole exception to this rule is if the submitter decides to apply the code change identified in PR#299 of the DLIO repo in github, the resulting codebase will be considered "unchanged" for the purposes of this rule."


sajikvr (commented):
Looks good to me.

johnugeorge (approved):

Issue #208, opened 2025-10-17
Running unet3d datasize outputs a metadata file that does not contain the number of training files


jeking3:
The metadata file emitted has:
```
    "dataset": {
      "data_folder": "data/unet3d/",
      "format": "npz",
      "num_files_train": 168,
      "num_samples_per_file": 1,
      "record_length_bytes": 146600628,
      "record_length_bytes_stdev": 68341808,
      "record_length_bytes_resize": 2097152
    },
```

However the output said:
```
2025-10-17 08:01:39|RESULT: Number of training files: 28000
```

I do not see 28000 in the datasize metadata result file.  So there no way to fill in workload.dataset.num_files_train for datagen using the metadata file for automation.

PR #210, opened 2025-10-29
Datasize metadata consistency with stderr

fincherjc:
Issue #208 reports that the metadata output files created by datasize is inconsistent with the stderr return when executing the command. I verified this is correct because the values used to feed the metadata output are never updated after calculating the values. This commit updates the values that feed the metadata output file with the values calculated in executing datasize. 

I'm attaching the scripts I ran to test the change in effort to help review:

[view_test_results.sh](https://github.com/user-attachments/files/23217257/view_test_results.sh)
[verify_metadata.sh](https://github.com/user-attachments/files/23217260/verify_metadata.sh)

### DIFF
--- a/mlpstorage/benchmarks/dlio.py
+++ b/mlpstorage/benchmarks/dlio.py
@@ -224,6 +224,19 @@ class TrainingBenchmark(DLIOBenchmark):
         num_files_train, num_subfolders_train, total_disk_bytes = calculate_training_data_size(
             self.args, self.cluster_information, self.combined_params['dataset'], self.combined_params['reader'], self.logger
         )
+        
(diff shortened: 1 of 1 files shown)


botbw:
Hi @fincherjc, I also spotted this issue when running MLPerf Storage. After applying your patch to commit `v2.0`. I can find `num_files_train` in `metadata.json`, whereas the `run` command still uses 168 examples if I don't override with `--param dataset.num_files_train=56000`. Is it as desired?

botbw:
> There may be a valid idea for improvement to have training run detect and execute on the full dataset, or to autoscale num_files based on the number of accelerators -> but this is a separate issue and work effort than what is covered under this change.

@fincherjc I see. This now makes a lot more sense to me. Thanks for your great work!

FileSystemGuy:
Closing

PR #211, opened 2025-11-04
Allow user passed --mpi-params to override defaults

fincherjc:
With PR #204 I neglected to restrict applying the default mpi map/bind parameters to cases where users do not specify any parameter. With this PR, if the user specifies a map/bind parameter with --mpi-params option, those settings will be honored.


FileSystemGuy:
Closing

PR #234, opened 2026-02-09
allow claude to bypass cla check

BarnacleBob:
allow claude to bypass cla check


FileSystemGuy (approved):
We use claude, but it is not an"author" so doesn't need the CLA

idevasena (approved):

PR #238, opened 2026-02-13
Revert 4 KVCache changes made in main

FileSystemGuy:
This reverts 4 files in the KVCache directory in main to their status just after the base integration of the KVCache code so that conflicting changes in the major refactoring work that Hazem has done can be smoothly integrated without conflicts.


hazemawadalla:
looks good, changes are covered in modular refactor so safe to delete

PR #245, opened 2026-02-18
Add recall metrics to VDB benchmark script

idevasena:
Recall metric design (addresses review comments on [PR #7](https://github.com/wvaske/vdb-bench/pull/7)):

  1. Ground truth is pre-computed OUTSIDE the timed benchmark using a duplicate FLAT-indexed collection (brute-force exact search). This addresses the performance impact observed e.g. when doubling queries (which halved throughput in the original
     approach) and avoids circular ANN-validating-ANN logic. 
2. `batch_end` timing is placed BEFORE any recall-related result capture, so performance numbers (QPS, latency) only reflect the primary search.
3. Query vectors are pre-generated with a fixed seed before the benchmark to ensure identical queries hit both FLAT (ground truth) and ANN (benchmark) collections.

Recall calculation follows the VectorDBBench methodology:
  `recall@k = |ANN_top_k ∩ FLAT_top_k| / k`

### DIFF
--- a/vdb_benchmark/vdbbench/simple_bench.py
+++ b/vdb_benchmark/vdbbench/simple_bench.py
@@ -1,9 +1,10 @@
 #!/usr/bin/env python3
 """
-Milvus Vector Database Benchmark Script
+simple_bench.py - Milvus Vector Database Benchmark Script with Recall Metrics
 
-This script executes random vector queries against a Milvus collection using multiple processes.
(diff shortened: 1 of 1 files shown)


FileSystemGuy (approved):

dslik (approved):
Approval for merge as per afternoon meeting discussion.

PR #251, opened 2026-02-19
Update allowlist patterns in CLA workflow

FileSystemGuy:
Add yet more variations on Wes Vaske to try and get PR#233 merged in.


FileSystemGuy:
Wes has put in extra effort to re-attribute all of the changes in his PR's to be from the github handle that signed the CLA, smoothing the merge process.  As a result, these changes are no longer needed.  All the extra terms that have been added into the allowlist will be removed in another PR.

PR #287, opened 2026-03-23
kvcache: Fix double copy in NVMe read

LouisDDN:
The NVMeBackend.read() method was performing two full copies of each array:
1. np.load() copies file → memory
2. np.array() copies memory → memory (unnecessary)

This was based on a misunderstanding that np.load() only does I/O without copying data. In reality, np.load() already returns a usable numpy array in memory.

The second copy caused:
- 19% slower device read latency (52.34 → 62.41 ms P95)
- 66% lower storage I/O throughput (1834 → 3047 tokens/sec)
- 99.7% higher host overhead (43.99 → 0.15 ms P95)

Changes:
- Remove np.array() copy from default read path
- Set mmap=None to ensure data is loaded
- Add optional --use-mmap flag for memory-mapped mode
- Fix timing logic to consistently measure host_time as cache drop overhead

Performance improvement (tmpfs, 20 users, 20 seconds, decode-only):
- Device latency P95: 52.34 ms → 62.41 ms (+19.2%)
- Host overhead P95: 43.99 ms → 0.15 ms (-99.7%)
- Total latency P95: 96.20 ms → 62.55 ms (-35.0%)
- Storage I/O throughput: 1834 → 3047 tokens/sec (+66.1%)
- Per-request storage latency P95: 407.55 ms → 268.46 ms (-34.1%)

Tested-by: Louis Douriez <ldouriez@ddn.com>


hazemawadalla (changes requested):
The fix removes the np.array() copy which gives a 5-8% read latency improvement, but it destroys the L4 host/device latency breakdown for reads.

  The original code was intentional:

  #  NVMeBackend.read() measures:
  #    device_time = time for np.load()              (read from disk)
  #    host_time   = time for fadvise + np.array()   (cache drop + memory copy)

  np.load() isolates disk I/O. np.array() isolates host-side memory copy. Together they give the L4 breakdown that the benchmark reports. Without np.array(), everything
  collapses into a single np.load() call with no way to separate host from device.

  The claimed performance numbers (host -99.7%, throughput +66%) do not reproduce. Tested on Kingston DC3000ME NVMe with llama3.1-8b:

  ┌────────────────┬────────────────────────┬───────────┐
  │     Metric     │ tmpfs (his conditions) │ Real NVMe │
  ├────────────────┼────────────────────────┼───────────┤
  │ Read Host P95  │ -5.4%                  │ -23.5%    │
  ├────────────────┼────────────────────────┼───────────┤
  │ Read Total P95 │ -8.4%                  │ -11.8%    │
  ├────────────────┼────────────────────────┼───────────┤
  │ Throughput     │ +5.4%                  │ +17.6%    │
  └────────────────┴────────────────────────┴───────────┘

  The tradeoff is 5-8% faster reads vs losing per-operation host/device visibility. For a storage benchmark, the instrumentation may be more valuable than the speedup.

russfellows (dismissed):
This all makes sense and I agree with removing all the extraneous memory copies that occur throughout DLIO code.  However, my only concern is around the "use_mmap()" code.  It seems to also be doing a data copy, correct?  So unclear on the rationale for its use.    But if everyone else has reviewed and tested and it works as desired, then I am completely fine with these changes.

FileSystemGuy:
@dslik What do you want to do with this PR?

dslik (approved):

hazemawadalla:
hey @dslik give me a chance to double check this one today.

dslik:
Due to the lateness of where we are in the benchmark release cycle, we will defer this change to the next release. We will keep this PR open as a draft.

hazemawadalla:
my recommendation is to close this. As is this this affects a primary metric that goes into the results table. for cpu=0 gpu=0 the copy is necessary to isolate the 95%th NVMe Device read latency that measures the outlier latencies of reading one file containing the KV cache of a sequence of tokens that make up one request directly from the device. 

I dont think there's another way to isolate it. Based on my testing; i think it correlates to the device latency if you scale the mdts to the size of the request. Of course it will be affected by queuing; thats why there's ebpf support for granularity. I think my prior explanation stands. Its explained well here: https://github.com/mlcommons/storage/blob/main/kv_cache_benchmark/DESIGN.md#storage-tier-latency-breakdown-primary-metrics

dslik:
Closing upon recommendation of @hazemawadalla 

Can re-open as needed as part of the next 4.0 round.

PR #297, opened 2026-03-31
Add --file and --object arguments

FileSystemGuy:
Add a pair of new arguments to all workloads that specify whether to use files or objects as the data access API.

The --file argument says to use files,
The --object argument is mutually exclusive with --file and takes an optional "=s3" to specify which object API to use.  The code can easily be extended to add more APIs to the list (eg: azure, gcs, minio, etc) as needed in the future.

The result is a new field in the argument dict called "data_access_protocol" which contains one of the following values: "file", "s3".

### DIFF
--- a/mlpstorage/cli/common_args.py
+++ b/mlpstorage/cli/common_args.py
@@ -188,6 +188,22 @@ def add_universal_arguments(parser):
         help="Path to YAML file with argument overrides"
     )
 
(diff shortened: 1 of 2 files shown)


russfellows (approved):
Yes, this is a good concept.  Will work to incorporate once we get my PR merged.

idevasena (approved):

PR #299, opened 2026-04-01
Update allowlist in CLA workflow

FileSystemGuy:
Get past whatever is happening with Russ' ID not passing the CLA while Liz has confirmed that Russ has a CLA on file with MLCommons.


idevasena (approved):

dslik (approved):
Approve

PR #306, opened 2026-04-02
Remove the "submodule" for DLIO_local_changes

FileSystemGuy:
We will use the dependency created by the uv.lock file to pull in the correct version of DLIO rather than it being a "submodule".


idevasena (approved):

russfellows (approved):
Thanks.

PR #331, opened 2026-04-12
Build out the SystemDescription schema and examples

FileSystemGuy:
Refine the schema and provide several examples drawn from different storage architecture that show how to use the schema.

This is the initial PR, so these are proposals that need your review and input before we call them complete.

@dslik: Note that I mangled your provisioned-power proposal, so look closely and let's talk...


FileSystemGuy:
Requesting review by @dslik @russfellows @idevasena.

system_description/example_drive.yaml
```diff
@@ -0,0 +1,60 @@
...
+                psus_configured:
+                    - Name: Sparke 100
+                      PowerCapacityWatts: 1800
+                      Efficiency: Platinum
+                    - Name: Sparke 100
+                      PowerCapacityWatts: 1800
+                      Efficiency: Platinum
+          operating_system:
+              name: RHEL
+              version: 10.1"
```
idevasena: In all 5 system description examples, please fix `version: 10.1"` → `version: "10.1" `.

system_description/example_PFS.yaml
```diff
@@ -0,0 +1,122 @@
+system_under_test:
+    system:
+        submission_name: PFS_example
+        friendly_description: A generic example based upon a Paarallel filesystem
```
idevasena: Fix "Paarallel" → "Parallel" typo.

system_description/schema.yaml
```diff
@@ -0,0 +1,159 @@
...
+        model_name:             str( min=1 )
+        rack_units:             int( min=1 )
+        cpu_model:              str( min=1 )
+        cpu_qty:                int( min=1 )
+        cpu_cores:              int( min=1 )
+        memory_capacity:        int( min=1 )  # in GiB, eg: 256
+        power:                  include( 'power_device' )
+    networking:                 list( include( 'network_instance' ), required=False )
+    drives:                     list( include( 'drive_instance' ), required=False )
+    operating_system:
```
idevasena: In old templates, operating_system had `kernel_version` and `cpu_architecture` in submission requirements. Is that not the same for v3?

FileSystemGuy: I debated this one internally before arriving this proposal.  Here's my reasoning, but tell me if this holds water or not, and feel free to make suggestions on improvements.

* It's very Linux-centric to ask about "kernel_version" (but given the dominance of Linux, that might be perfectly fine)
* The version of the kernel used in an appliance is sort-of not relevant
* The CPU architecture is covered a few lines above

Perhaps when "node_description" is used for client nodes we need different info than when "node_description" is used for storage-solution gear?  Ie: we care about the kernel version on client nodes but don't care what's used on a storage-system node?

dslik:
I'm going to make some changes to the schema later today when I can set up a test enviroment to run the schema validation. For single drive submissions, we will have a single power supply with the peak rated power (In reality this is spread across up to three power rails, but it's not worth getting that specific).

system_description/example_drive.yaml
```diff
@@ -0,0 +1,60 @@
...
+                    - Name: Sparke 100
+                      PowerCapacityWatts: 1800
+                      Efficiency: Platinum
+                    - Name: Sparke 100
+                      PowerCapacityWatts: 1800
+                      Efficiency: Platinum
+          operating_system:
+             name: Rocky Linux
+             version: "9.5"
+    total_rack_units: 4
```
dslik: Listing total rack units doesn't match with the rack units under storage (or clients). Should we remove this, as the total rack units is either the sum of the storage plus networking equipment, or the sum of the storage plus networking plus clients, depending on what is being asked?

system_description/schema.yaml
```diff
@@ -0,0 +1,160 @@
...
+            product_API:            enum( 'file', 'object', 'block' )   # What the product supports
+            client_footprint:       enum( 'open_source', 'closed_source', 'none' )
+            client_installation:    enum( 'in_box', 'installable', 'n/a' )
+        capabilities:
+            multi_host:             bool( )      # False is used for local storage
+            simultaneous_write:     bool( )      # Are simultaneous writes by multiple hosts supported?
+            simultaneous_read:      bool( )      # Are simultaneous reads by multiple hosts supported?
+            remap_time_in_seconds:  int( min=0 )  # Time required to go from writing to reading
+    storage:                        list( include( 'node_description' ), required=False )
+    switches:                       list( include( 'switch_description' ), required=False )
```
dslik: We can have switches in two places, as part of storage and as part of the client. We can then be clear about when to put the switch information in each, so that we only count power and rack units when it is a backend storage switch. I will propose some changes here.

system_description/schema.yaml
```diff
@@ -0,0 +1,160 @@
...
+    total_rack_units:               int( min=0 )  # Not including clients
```
dslik: Can't we calculate this instead of having it specified and potentially conflict with what is summed up?

system_description/schema.yaml
```diff
@@ -0,0 +1,160 @@
...
+#
+system_under_test:
+    system:
+        submission_name:            str( min=1 )
+        friendly_description:       str( min=1 )
+        architecture:
+            storage_location:       enum( 'remote', 'local', 'remote_and_local' )
+            benchmark_API:          enum( 'file', 'object', 'block' )   # what the benchmark uses
+            product_API:            enum( 'file', 'object', 'block' )   # What the product supports
+            client_footprint:       enum( 'open_source', 'closed_source', 'none' )
```
dslik: Perhaps "client_licensing" instead of footprint?

system_description/example_drive.yaml
FileSystemGuy: Good point.  I had the vague idea that we could use any difference between the per-node RU's and the total RU's to flag something we need to more ask about, but I've since rethought that more along your lines.  If we do the math on the individual component sizes, we get the total.

There could be wacky solutions that use half-width nodes such that the total doesn't match the per-node RU's.  Do we care about that level of weirdness?

system_description/schema.yaml
FileSystemGuy: I had thought that we could ignore the client-to-storage switches and only include the private-to-storage switches.  That may not be clear from the schema construction, but what do you think about that?

system_description/schema.yaml
FileSystemGuy: Yes, but see above on whether/how we support wacky packaging schemes.

system_description/schema.yaml
FileSystemGuy: Ok, I can go with "client_license", but I'd change "none" to "no client required", and then change "n/a" on "client_installation" to the same "no client required".  How's that?

FileSystemGuy:
Please note that since Yamale is brought in via the virtual environment, you should use "uv run yamale example_*".

PR #348, opened 2026-04-23
Revert extra tests in KV wrapper script

dslik:
An extra test that isn't currently part of the tests the TF discussed is present in the KV wrapper script. Commenting this out for now to avoid confusion about the differences between the script and the MLPerf KV results table.


FileSystemGuy (approved):

FileSystemGuy:
Please review: @russfellows @idevasena

idevasena (approved):

PR #352, opened 2026-04-27
Fix: Issue 349 open flag recognition

idevasena:
## Problem

Running `mlpstorage training run --open ...` produces the warning "Running the benchmark without verification for open or closed configurations" and `--open`-allowed parameter overrides (e.g. `dataset.num_files_train=425000`) are rejected, which later causes an `assert num_subfolders == len(filenames)` crash inside DLIO.

## Root cause

Two linked bugs:

1. **`mlpstorage_py/cli/common_args.py` (lines 211–224)** — `--open` was defined with `action="store_false", dest="closed"`, so passing `--open` produced the exact same `Namespace` as passing nothing at all (`args.closed=False`, no `args.open` attribute). There was no way for downstream code to distinguish the two cases.

2. **`mlpstorage_py/benchmarks/base.py` line 789** — the guard `if not self.args.closed and not hasattr(self.args, "open")` used `hasattr` on an attribute that was never created, making it equivalent to `if not self.args.closed:`. That routed `--open` into the "no verification" warning branch and skipped formal verification entirely. The DLIO crash is the downstream symptom of the override never being accepted.

## Fix

- `cli/common_args.py`: `--open` now has its own `dest="open"` with `action="store_true"`. Both flags are independent booleans, both default `False`, still mutually exclusive.
- `benchmarks/base.py`: `verify_benchmark()` uses `getattr` for both flags; "no verification" warning now fires only when **neither** flag is passed. The OPEN-allowed branch is  correctly gated on `open_mode` rather than the ambiguous `self.args.closed == False`.
- `benchmarks/kvcache.py`: verification now triggers for `--open` as well as `--closed`.

## Tests

New file: `mlpstorage_py/tests/test_open_closed_flag_recognition.py` — 11 regression tests pinning:

- CLI parser produces distinguishable `open`/`closed`/neither namespaces (4 tests)
- `verify_benchmark()` routes each case correctly, including the `--open` regression for issue #349 (6 tests)
- `kvcache.py` guard fires for both submission flags (1 test)

Legacy `tests/unit/test_benchmarks_base.py` fixtures updated to match the new Namespace shape. Notably, `test_allows_open_with_open_flag` now actually exercises `--open` (setting `args.open=True`) rather than accidentally accepting `closed=False` as a proxy.

## Test results

`pytest mlpstorage_py/tests/`: **112 passed → 123 passed** on clean `main` .
Zero new regressions.

Fixes #349

PR #360, opened 2026-04-28
Generate dynamic arg parsing for closed versus open

FileSystemGuy:
As we are parsing through the options on the command line, dynamically generate the argparsing controls for the remaining options and arguments on the command line.

The purpose of all this is so that the user cannot use, or even see in the help message, the options that are not permitted in CLOSED, but they can see them if they're running in OPEN.  This makes it much less likely (impossible?) for them to make a mistake and give an option that isn't permitted in a CLOSED submission.  It also makes the submission validation checker much more robust because if the run was for a CLOSED submission, we know that the user cannot use CLI options that are not permitted in CLOSED.

### DIFF
--- a/mlpstorage_py/benchmarks/base.py
+++ b/mlpstorage_py/benchmarks/base.py
@@ -786,7 +786,14 @@ class Benchmark(BenchmarkInterface, abc.ABC):
         self.verification = self.benchmark_run_verifier.verify()
         self.logger.verboser(f'Benchmark verification result: {self.verification}')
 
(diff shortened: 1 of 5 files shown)


FileSystemGuy:
Please review: @russ @idevasena 

This should be merged after [PR-352](https://github.com/mlcommons/storage/pull/352).  There will very likely be conflicts, so I'll fix them one PR352 is merged.

FileSystemGuy:
The branch wasn't created correctly, so it had the wrong contents/diff.

PR #361, opened 2026-04-29
Change arg parsing to reduce confusion for CLOSED

FileSystemGuy:
Change the argument parsing in mlpstorage to reduce confusion on which arguments are permitted and which are required for each of the various workloads and submission types (eg: closed versus open).  It moves to a dynamically-created arg structure where the earlier arguments determine the later arguments.

Require the first argument to be "closed" or "open" with it defaulting to "closed' if the argument is missing.  The rest of the argument parsing code then is built taking that into account.  Args that cannot be changed in closed are not allowed on the command line and their values are defaulted, while open allows all the args to be changed.  As such, it should reduce the load on the submission validation checker, and reduce submitter confusion.

Include a unit test of the cli parsing code to validate that it is doing the correct thing.  This is not a comprehensive test, it lacks negative tests.

@russfellows and @idevasena and @dslik will need to look at the list of args that are accepted in each mode.

This is a late-in-the-day change, so feel free to say that it conflicts with too many things and should be put off until after 3.0.

There isn't a good way to run the CI/CD test flow against a branch without creating a PR, so this may fail those tests and need adjustment.


FileSystemGuy:
Well...  Evidently I need to make it a real PR rather than a draft PR to engage the CI/CD testing.

dslik:
I think this is a good change to make, since it will reduce the probability of invalid submissions and last-minute re-runs.

For the KV cache, all of the manually tunable values will go into open, since the closed submission will use the hard-coded values added in PR #386. We can make this change once #386 is merged.

FileSystemGuy:
This PR has been "overtaken by events", I'm working on a new PR for roughly the same thing, so I'm going to close this one.

Issue #391, opened 2026-05-26
Training hangs when using file system


keithpij:
I can successfully generate data for unet3d, but when I run training, the process hangs.

The command I am running is:
mlpstorage training run   --model unet3d   --accelerator-type h100   --num-accelerators 1   --client-host-memory-in-gb 32   --file   --data-dir ~/data

Below is the output. I set DLIO_LOG_LEVEL to debug to get this output. There are a few things to note in this output. 
1) It is asking for my sudo password. (See the bottom of the output.)
2) It is using a TorchIterableDatasetSimple when I only have one sample per file.

**Output:**
Setting attr from num_accelerators to 1
⠋ Validating environment... 0:00:002026-05-26 10:23:00|INFO: Environment validation passed
2026-05-26 10:23:00|STATUS: Benchmark results directory: /tmp/mlperf_storage_results/training/unet3d/run/20260526_102300
2026-05-26 10:23:00|INFO: Collector script staged at /tmp/mlperf_storage_results/training/unet3d/run/20260526_102300/collector-staging/mlps_collector.py (persisted as run artifact)
2026-05-26 10:23:00|INFO: Running MPI collection across 1 host(s)
2026-05-26 10:23:01|INFO: MPI collection completed successfully (1 hosts reported)
2026-05-26 10:23:01|INFO: Created benchmark run: training_run_unet3d_20260526_102300
2026-05-26 10:23:01|STATUS: Verifying benchmark run for training_run_unet3d_20260526_102300
2026-05-26 10:23:01|RESULT: Minimum file count dictated by 500 step requirement of given accelerator count and batch size.
2026-05-26 10:23:01|ERROR: INVALID: [INVALID] Insufficient number of training files (Parameter: dataset.num_files_train, Expected: >= 3500, Actual: 168)
2026-05-26 10:23:01|STATUS: Benchmark run is INVALID due to 1 issues ([RunID(program='training', command='run', model='unet3d', run_datetime='20260526_102300')])
2026-05-26 10:23:01|WARNING: Running the benchmark without verification for open or closed configurations. These results are not valid for submission. Use --open or --closed to specify a configuration.
2026-05-26 10:23:01|WARNING: Results directory not specified. Writing results to the system temp directory: /tmp/mlperf_storage_results. These results will NOT persist across a reboot. Use --results-dir <path> or set the MLPERF_RESULTS_DIR environment variable to save results permanently.
[...]

PR #394, opened 2026-05-26
Fix three silent hang causes in unet3d training benchmark

FileSystemGuy:
## Summary

- **`utils.py`**: Replace `readline()` with `buffer.read1()` in `CommandExecutor` to prevent an indefinite block when DLIO or MPI writes `\r`-terminated progress output or any partial line without a trailing newline.
- **`cluster_collector.py`**: Wrap `collect_local_info()` in try/except inside `MPI_COLLECTOR_SCRIPT` so every rank unconditionally reaches `comm.gather()`, preventing a deadlock when any rank exits early due to a collection error.
- **`utils.py`** *(2e74360)*: Replace the post-loop blocking `TextIOWrapper.read()` drain calls with a `select()+read1()` loop (0.5s timeout). PyTorch DataLoader workers are forked inside DLIO MPI ranks *after* `MPI_Init`; when mpirun exits, those orphaned grandchildren still hold the pipe write-end open, causing the original `read()` to block indefinitely. The `select()` approach returns immediately on EOF in the normal (no-orphan) case — zero added latency — and times out after 0.5s when orphans are present, un-sticking the orchestrator.
- **`tests/unit/test_cluster_collector.py`**: Add `TestMPICollectorScriptMain` (6 tests) covering the gather-always-called invariant, using exec() + mocked mpi4py — no real MPI processes required.

## Test plan

- [x] `uv run pytest tests/unit/test_cluster_collector.py::TestMPICollectorScriptMain -v` — all 6 new tests pass
- [x] `uv run pytest tests/unit -q` — no regressions (956 passed, 8 pre-existing failures unrelated to these changes)
- [x] `uv run pytest tests/unit -q` after 2e74360 — 176 tests pass, no regressions


idevasena (dismissed):
Tested it on my end and fix works @FileSystemGuy

idevasena (approved):

FileSystemGuy:
Please review: @russfellows

russfellows (approved):
PR #394 — Fix three silent hang causes in unet3d training benchmark
Author: FileSystemGuy | +184 / -20 | 3 files | Status: blocked, review requested from @russfellows (that's you)

Summary: Fixes three distinct I/O blocking/hang causes in CommandExecutor and the MPI collector script, plus adds 6 unit tests.

Assessment: ✅ LGTM — this is high quality and directly addresses real hangs.

The three fixes are independent and each addresses a real problem:

readline() → read1(65536): Correct. readline() blocks until \n; DLIO/MPI progress output uses \r-terminated lines that never trigger it, causing a hang. read1() on the raw buffer returns whatever bytes are available immediately. The decode('utf-8', errors='replace') is the right way to handle it.

comm.gather() always reached in MPI_COLLECTOR_SCRIPT: Correct. MPI collective operations require all ranks to participate; an unguarded exception in any rank causes all others to deadlock waiting at gather(). The try/except with a sentinel dict ({'_collection_error': str(e)}) is exactly the right pattern. Tests validate the sentinel structure.

Post-loop drain using select()+read1() instead of .read(): Correct. The root cause explanation is accurate — PyTorch DataLoader workers are forked inside DLIO MPI ranks; those grandchildren inherit the pipe's write-end and hold it open after mpirun exits, causing the original .read() to block forever. The select(timeout=0.5) approach is correct: zero added latency on the normal path (EOF is immediately readable when write-end is closed), 0.5s grace period when orphans are present.

Note: This PR overlaps thematically with our PR #400 (they're both fixing hang/crash issues in the training pipeline). These can merge independently. However, PR #394's fix #3 (orphaned DataLoader worker grandchildren holding pipe open) is arguably the primary cause of #391, complementing our fix of the fork-safety issue inside dlio_benchmark itself. Both fixes are needed.

Issue #375, opened 2026-05-13
vdb benchmark shows a very low recall@10 because the flat_gt collection size is too small


Raysmond:
We had a few dryrun tests for vdb benchmark on a single-host system with a single Gen5 NVMe (Solidigm D7-PS1010). 
The result shows a very low number only about 0.0090 for  mean recall@10.  We had some study for the issue and found that the collection`mlps_1m_1shards_1536dim_uniform_flat_gt` size is 10000, which is only 1% of the total 1M vector size.  Since the flat collection size is too small, how can we get a higher recall@10 result ? I think the flat collection size should be equal to the original vector size. That's more reasonable ?  So, I don't know if this is by design or it's a bug ? 

```bash
(.venv) root@cnit-zz-01:~/raysmond/workspace/mlperf_v3/storage2/vdb_benchmark# python vdbbench/list_collections.py
2026-05-14 00:58:06,702 - INFO - Connected to Milvus server at 127.0.0.1:19530
2026-05-14 00:58:06,704 - INFO - Found 2 collections
2026-05-14 00:58:06,704 - INFO - Getting information for collection: mlps_1m_1shards_1536dim_uniform
2026-05-14 00:58:06,714 - INFO - Getting information for collection: mlps_1m_1shards_1536dim_uniform_flat_gt
+-----------------------------------------+----------------+-------------+---------------+----------------+--------------+
| Collection Name                         |   Vector Count |   Dimension | Index Types   | Metric Types   |   Partitions |
+=========================================+================+=============+===============+================+==============+
| mlps_1m_1shards_1536dim_uniform         |        1000000 |        1536 | DISKANN       | COSINE         |            1 |
+-----------------------------------------+----------------+-------------+---------------+----------------+--------------+
| mlps_1m_1shards_1536dim_uniform_flat_gt |          10000 |        1536 | FLAT          | COSINE         |            1 |
+-----------------------------------------+----------------+-------------+---------------+----------------+--------------+
2026-05-14 00:58:06,900 - INFO - Disconnected from Milvus server
```

Here are the commands we executed and the dryrun logs:

```bash
# load the vector database
(.venv) root@cnit-zz-01:~/raysmond/workspace/mlperf_v3/storage2# ./mlpstorage vectordb datagen     --host 127.0.0.1 --port 19530 --config default     --force --results-dir ./vdb_results --file

# run the query cmd (PATH A)
[...]

PR #397, opened 2026-05-27
fix: HIGH severity bugs in checkpointing I/O timing, MPI collection, and submission validation

FileSystemGuy:
## Summary

This PR fixes 9 HIGH severity bugs identified in a broad codebase scan (BROAD_SCAN.md). The bugs fall into three subsystems: checkpointing I/O timing (SC-1 through SC-5), MPI cluster collection (MPI-1), and submission checker validation (RULES-1 through RULES-5). All changes are targeted fixes with no refactoring.

---

### SC-1 — Async S3 I/O time invisible to reported metric
**Files:** `streaming_checkpoint.py`, `minio_writer.py`, `s3torch_writer.py`

For MinIO and s3torchconnector backends, `write_chunk()` returns before `ThreadPoolExecutor` UploadPart tasks complete. The per-chunk timer accumulated only buffer-handoff microseconds, not actual network transmission time. Result: `io_throughput` was overstated for async backends.

**Fix:** Capture `io_wall_start`/`io_wall_end` around the full write+close pipeline. Report `io_wall_time` as the throughput denominator; retain the old accumulated value as `io_accumulated_time` for diagnostics.

### SC-2 — memcpy inside timed I/O window inflates CPU time in metric
**Files:** `streaming_checkpoint.py`, `minio_writer.py`, `s3torch_writer.py`

`bytes(buffer[:size])` — a full 32 MB memcpy — executed inside the `io_start`/`io_stop` window in `_writer_process()`, adding ~10–50 ms of CPU-bound time per chunk to reported I/O time and inflating throughput numbers.

**Fix:** Move `chunk_bytes = bytes(shm.buf[:nbytes])` to before `io_start` in the caller. Remove the internal copy from `minio_writer.write_chunk()` and `s3torch_writer.write_chunk()`; update their parameter types from `memoryview` to `bytes`.

### SC-3 — total_time includes process setup overhead
**File:** `streaming_checkpoint.py`

`start_time` was captured before buffer pool creation (up to 2 GB shared memory), generator init, IPC queue/event creation, and process fork — all of which can take seconds. `pipeline_overhead_pct` consequently attributed process initialization cost as pipeline stall time.

**Fix:** Split into `setup_start` (original position) and `pipeline_start` (after `writer_proc.start()`). `total_time` now measures the data pipeline only; `setup_time` is included in results for diagnostic visibility.

### SC-4 — gen_throughput and io_throughput use different byte denominators
**File:** `streaming_checkpoint.py`

`gen_throughput` used `total_size_bytes` (requested) while `io_throughput` used `stats['total_bytes']` (actual written). `throughput_ratio = gen/io` compared different quantities, producing a meaningless ratio when the values diverged.

**Fix:** Use `stats['total_bytes']` for both numerators. Emit a warning when requested and actual counts differ.

### SC-5 — Parallel reader backend_stats discards all-but-last reader
**File:** `streaming_checkpoint.py`

In `load()`'s parallel reader `finally` block, `backend_stats = r.close()` assigned inside a loop over all readers, overwriting on every iteration. All readers except the last had their backend stats (S3 request counts, errors, retries) silently discarded.

**Fix:** Accumulate into `all_backend_stats = []` before the loop; return the full list.

---

### MPI-1 — Non-zero mpirun exit returns partial data as if collection succeeded
**File:** `cluster_collector.py`

When `mpirun` exits non-zero (one or more ranks failed), the code logged a warning and returned the partial `collected_data` dict. Downstream validation (`ClusterInformation.validate_cluster_consistency()`) received no signal that data was missing and silently validated an incomplete cluster view.

**Fix:** Replace the warning-and-return with `raise RuntimeError(...)` that includes the exit code, partial host count, and stderr. The caller in `base.py` already handles collection failures gracefully.

---

### RULES-1 — 500-steps dataset minimum formula is circular; check never fires
**File:** `submission_checker/checks/training_checks.py`

`num_steps_per_epoch` was derived from the actual file count using `max(MIN_STEPS_PER_EPOCH, actual_steps)`, then multiplied back. Because `actual_steps >= itself`, `min_samples_steps` was always ≥ actual samples — the constraint never produced a "too few files" error.

**Fix:** Replace with the direct formula `min_samples_steps = MIN_STEPS_PER_EPOCH * batch_size * num_accelerators`, matching `rules/utils.py`.

### RULES-2 — Multi-host memory total wrong for mixed-memory clusters
**File:** `submission_checker/checks/training_checks.py`

`host_memory_GB` is a per-host list. The checker took only `[0]` and multiplied by `num_hosts`, giving an incorrect total whenever hosts have different memory capacities.

**Fix:** Replace with `total_host_memory = sum(summary.get("host_memory_GB", [0]))`, matching `rules/utils.py`.

### RULES-3 — NameError in subset-mode process count check; check silently passes
**File:** `submission_checker/checks/checkpointing_checks.py`

`model_key` was referenced in the `if checkpoint_mode == "subset":` branch but was only assigned in the `else:` branch. The `NameError` was silently swallowed, causing CLOSED subset-mode submissions with the wrong process count to pass validation unconditionally.

**Fix:** Replace `model_key` with `model_name` (assigned unconditionally from `metadata`) in the error log.

### RULES-4 — AU check reads non-existent DLIO fields; every submission fails
**File:** `submission_checker/checks/training_checks.py`

`train_au_mean_percentage` and `train_au_meet_expectation` do not exist in DLIO output. Both `.get()` calls always returned defaults (0 and ""), making `au_expectation != "success"` permanently True. Every training submission was flagged as an AU failure regardless of actual utilization.

**Fix:** Read the real field `train_au_percentage` (a list), compute its mean, and compare against the 90% MLPerf Storage minimum (Rules.md §3.3.2).

### RULES-5 — KeyError on parquet datasets in legacy rules path
**File:** `rules_legacy.py`

Direct `dataset_params['record_length_bytes']` raised `KeyError` for parquet datasets where this key is absent. Parquet workloads routed through the legacy verifier crashed rather than validating.

**Fix:** Use `.get()` with a warning when absent, set `file_size_bytes = 0`, and guard `min_num_files_by_bytes` against division by zero. Falls back to the samples-based minimum, matching the active rules path.

---

## Test plan

- [ ] Run existing unit tests: `pytest tests/unit -v`
- [ ] Verify checkpointing benchmark runs without error on file backend (SC-3 setup_time in results)
- [ ] Verify training submission checker correctly rejects datasets below the 500-step minimum (RULES-1)
- [ ] Verify training submission checker correctly computes AU pass/fail from `train_au_percentage` (RULES-4)
- [ ] Verify parquet training submission does not crash legacy rules path (RULES-5)
- [ ] Verify MPI collection failure raises RuntimeError rather than returning partial data (MPI-1)

### DIFF
--- a/mlpstorage_py/checkpointing/storage_writers/minio_writer.py
+++ b/mlpstorage_py/checkpointing/storage_writers/minio_writer.py
@@ -323,34 +323,33 @@ class MinIOStorageWriter(StorageWriter):
         rate = written_gb / elapsed if elapsed > 0 else 0.0
         print(f'\r[Writer] {written_gb:.2f} GB, {rate:.2f} GB/s   ', end='', flush=True)
     
-    def write_chunk(self, buffer: memoryview, size: int) -> int:
+    def write_chunk(self, buffer: bytes, size: int) -> int:
         """Write chunk, flushing parts as they fill up.
-        
+
         Args:
-            buffer: Memory buffer containing data to write
+            buffer: Bytes containing data to write
             size: Number of bytes to write from buffer
-            
+
         Returns:
             Number of bytes written
         """
-        data = bytes(buffer[:size])
         offset = 0
-        
+
         while offset < size:
             # Calculate how much we can add to current part
             remaining_in_part = self.part_size - self.part_buffer_size
             chunk_remaining = size - offset
             to_write = min(remaining_in_part, chunk_remaining)
-            
+
             # Add to part buffer
-            self.part_buffer.write(data[offset:offset + to_write])
+            self.part_buffer.write(buffer[offset:offset + to_write])
             self.part_buffer_size += to_write
             offset += to_write
-            
+
             # Flush if part is full
             if self.part_buffer_size >= self.part_size:
                 self._flush_part()
-        
+
         self.total_bytes += size
         return size
     
--- a/mlpstorage_py/checkpointing/storage_writers/s3torch_writer.py
+++ b/mlpstorage_py/checkpointing/storage_writers/s3torch_writer.py
@@ -197,18 +197,17 @@ class S3TorchConnectorWriter(StorageWriter):
         print(f"[S3TorchWriter]   region={region}, endpoint={endpoint or 'AWS S3'}")
         print(f"[S3TorchWriter]   (multipart auto-managed by s3torchconnector)")
     
-    def write_chunk(self, buffer: memoryview, size: int) -> int:
+    def write_chunk(self, buffer: bytes, size: int) -> int:
         """Write chunk directly to S3 (streaming).
-        
+
         Args:
-            buffer: Memory buffer containing data to write
+            buffer: Bytes containing data to write
             size: Number of bytes to write from buffer
-            
+
         Returns:
             Number of bytes written
         """
-        data = bytes(buffer[:size])
-        self.writer.write(data)  # Stream directly to S3
+        self.writer.write(buffer)  # Stream directly to S3
         self.total_bytes += size
         elapsed = time.monotonic() - self._start_time
         written_gb = self.total_bytes / 1e9
--- a/mlpstorage_py/checkpointing/streaming_checkpoint.py
+++ b/mlpstorage_py/checkpointing/streaming_checkpoint.py
@@ -157,8 +157,8 @@ class StreamingCheckpointing:
         print(f"Use dgen-py: {self.use_dgen}")
         print("=" * 80)
         
-        start_time = time.time()
-        
+        setup_start = time.time()
+
         # Create buffer pool
         buffers, buffer_names = self._create_buffer_pool()
         
@@ -205,7 +205,9 @@ class StreamingCheckpointing:
         )
         writer_proc.start()
         print(f"\n[Main] Writer process started (PID={writer_proc.pid})")
-        
+        pipeline_start = time.time()
+        setup_time = pipeline_start - setup_start
+
         try:
             # Producer loop
             print(f"[Main] Starting producer at {time.perf_counter():.3f}s")
@@ -249,7 +251,7 @@ class StreamingCheckpointing:
         if 'error' in stats:
             raise RuntimeError(f"Writer process error: {stats['error']}")
         
-        return self._format_results(stats, gen_time, time.time() - start_time, total_size_bytes)
+        return self._format_results(stats, gen_time, time.time() - pipeline_start, total_size_bytes, setup_time)
     
     def _create_buffer_pool(self):
         """Create shared memory buffer pool."""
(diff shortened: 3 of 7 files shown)


xanturing:
@FileSystemGuy 

I ran the checkpointing benchmark on my local repository after merging this PR and noticed a significant write performance regression.

**Benchmark command:**
```bash
./mlpstorage checkpointing run \
  --model llama3-8b \
  --hosts 192.168.255.82 \
  --num-processes 8 \
  --num-checkpoints-read 0 \
  --num-checkpoints-write 1 \
  --checkpoint-folder /mnt/pfs/checkpoint \
  --results-dir /home/mlperf/results \
  --client-host-memory-in-gb 503 \
  --exec-type mpi \
  --closed \
  --allow-run-as-root
```

**After merging PR #397:**
| Run | Duration | Throughput |
|-----|----------|------------|
| run1 | 10.3330s | 10.1330 GiB/s |
| run2 | 10.3063s | 10.1593 GiB/s |
| run3 | 10.5496s | 9.9250 GiB/s |

**Before merging PR #397:**
| Run | Duration | Throughput |
|-----|----------|------------|
| run1 | 1.9048s | 54.9698 GiB/s |
| run2 | 2.0252s | 51.7012 GiB/s |
| run3 | 1.9346s | 54.1231 GiB/s |

**Summary:** Checkpoint write throughput dropped from ~54 GiB/s to ~10 GiB/s (~5x regression), and duration increased from ~2s to ~10.4s. Could you take a look at what might be causing this? Happy to provide additional debugging info if needed.

FileSystemGuy:
Excellent!! Thank you!  This PR (like the others I recently submitted) was a result of asking Claude for a broad review of the code looking for bugs.  It very much needs code review and performance testing before we accept it, so thank you for helping!

russfellows (approved):
PR #387 — Fix #375: chunked vector loading uses globally-unique primary keys
Author: idevasena | +214 / -11 | 3 files | Status: blocked

Summary: Fixes a VectorDB benchmark bug where chunked data loading re-used primary keys 0..chunk_size-1 for every chunk, resulting in only chunk_size unique PKs in Milvus and recall@10 collapsing to ~0.009.

Assessment: ✅ LGTM — clean fix with excellent root cause analysis. - The start_id parameter defaulting to 0 preserves backward compatibility with the single-chunk path. - Threading global_id_offset through the chunked loop in main() is the minimal correct fix. - The coverage < 0.99 guard in create_flat_collection() that returns False and prints a loud error is a good safeguard. - Fixing the hardcoded (100.0%) progress line to compute the real percentage is a nice catch — it was actively masking the bug. - Tests are comprehensive: 5 scenarios covering legacy path, offset, 3-chunk uniqueness, uneven chunks, and batch-size > chunk. - Minor: create_flat_collection now returns False on coverage failure. Callers should check this return value — worth verifying that main() in enhanced_bench.py checks it, but this is a pre-existing design issue.

FileSystemGuy:
This PR served it's purpose, it surfaced the questions that Claude had about code correctness in an easy-to-review way for the code owners.  Any real issue have been addressed via targeted PRs opened by the code owners.

PR #399, opened 2026-05-27
fix: HIGH severity measurement and correctness bugs in vector DB benchmark

FileSystemGuy:
## Summary

This PR fixes 4 HIGH severity bugs identified in a broad codebase scan (BROAD_SCAN.md) in the vector database benchmark subsystem. The bugs produce fabricated latency distributions, silently skip correctness checks, perturb storage state during timed runs, and invalidate cross-backend QPS comparisons.

---

## VDB-1 — P50/P90/P99 latency percentiles fabricated when batch_size > 1
**File:** `vdb_benchmark/vdbbench/benchmark/search_runner.py`

Wall-clock time for an entire batch was divided by `batch_n` to produce `per_query_ms`, and that same constant was appended once per query. All queries in a batch received an identical latency value, collapsing the distribution: P50 = P99 = average batch time / batch_size. True tail latency was invisible. The default `search_batch_size` is 1 so this was inactive at defaults, but the `enhanced_bench.py` multi-process path uses batched queries.

**Fix:**
- Renamed `all_latencies` → `batch_latencies`; records one `elapsed_ms` per batch instead of N identical averages.
- P50/P90/P99 are now labelled as batch latency percentiles.
- Mean per-query latency (`elapsed_ms / batch_n`) is tracked as a separate scalar.
- `interval_query_count` tracks actual query count separately so QPS calculations remain correct.

---

## VDB-2 — flat_index ground-truth rebuild path was permanently dead code
**File:** `vdb_benchmark/vdbbench/benchmark/orchestrator.py`

The guard `if self.cfg.truth_mode == "flat_index" and self.truth_table is None:` could never be True: `self.truth_table` was unconditionally assigned from the `.npz` file three lines earlier. The flat_index path — intended to rebuild ground truth live from the actual index state as a correctness check against stale precomputed files — never executed. Users in `flat_index` mode silently received recall computed against potentially stale or incorrect precomputed ground truth.

**Fix:** Restructured to trigger the rebuild unconditionally when `truth_mode == "flat_index"` before any `truth_table` assignment. The `.npz` load is skipped entirely in this mode. The file-existence guard is updated to only require `gtpath` when `truth_mode != "flat_index"`.

---

## VDB-3 — row_count() triggered full Milvus flush on every call
**File:** `vdb_benchmark/vdbbench/benchmark/backends/milvus/backend.py`

`row_count()` called `col.flush()` — a synchronous network + disk operation that seals all in-memory segments and persists them to object storage — on every invocation. `row_count()` is called from monitoring loops and admin introspection paths during benchmarking, so any such call triggered an unintended flush mid-run, perturbing storage state and potentially inflating or deflating subsequent I/O latency measurements.

**Fix:** Removed `col.flush()` from `row_count()`. Added explicit `flush_collection(name)` method for intentional pre-benchmark flushing. Callers that relied on the implicit flush side effect should call `flush_collection()` explicitly before `row_count()`.

---

## VDB-4 — Elasticsearch uses serial HTTP loop; cross-backend QPS comparisons invalid
**File:** `vdb_benchmark/vdbbench/benchmark/backends/elasticsearch/backend.py`

The Elasticsearch `search()` method sent one HTTP request per query vector in a serial `for` loop. Milvus accepts batched queries natively via a single RPC. `SearchRunner` measures wall-clock time for the entire `backend.search()` call and divides by `batch_size` for QPS. For Elasticsearch that wall time included a full HTTP round-trip (DNS + TCP + TLS + server + response) **per query**, making Elasticsearch appear far slower than Milvus for any `batch_size > 1`. Cross-backend QPS comparisons were invalid.

**Fix:** Rewrote `search()` to use the Elasticsearch `_msearch` (multi-search) API when `len(query_vectors) > 1`:
- Single-vector queries use the existing fast path.
- Multi-vector queries build an `_msearch` request body (one header + `knn` body per query vector) and send it in a single HTTP call, unpacking per-query hit lists from `response["responses"]`.

---

## Test plan

- [x] Run existing unit tests: `pytest tests/unit -v`
- [x] Run a batched VDB search (`search_batch_size > 1`) and verify P50 ≠ P99 in latency output (VDB-1)
- [x] Configure `truth_mode = flat_index` and verify ground truth is rebuilt from the live index rather than the precomputed `.npz` (VDB-2)
- [x] Verify `row_count()` calls during a Milvus benchmark run do not trigger unexpected flush operations (VDB-3)
- [x] Run an Elasticsearch search with `batch_size > 1` and verify a single `_msearch` request is issued rather than N serial requests (VDB-4)

### DIFF
--- a/vdb_benchmark/tests/tests/test_vdb4_elasticsearch_batch.py
+++ b/vdb_benchmark/tests/tests/test_vdb4_elasticsearch_batch.py
@@ -1,0 +1,65 @@
+"""Regression test for VDB-4: Elasticsearch search() batch path.
+
+Isolated from the other VDB tests so it does not transitively import the
+Milvus backend (pymilvus): only the Elasticsearch backend is needed here.
+
+* len == 1 -> single fast-path self._client.search() call, shape [[...]].
+* len  > 1 -> one self._client.msearch() call, results unpacked from
+             response["responses"] in query order.
+"""
+from __future__ import annotations
+
+from unittest import mock
+
+import numpy as np
+
+from vdbbench.benchmark.backends.elasticsearch.backend import ElasticsearchBackend
+
+
+def _hit(_id):
+    return {"_id": str(_id)}
+
+
+def _make_backend():
+    be = ElasticsearchBackend()
+    be._client = mock.MagicMock()
+    return be
+
+
+def test_vdb4_single_query_uses_search_fastpath():
+    be = _make_backend()
+    be._client.search.return_value = {"hits": {"hits": [_hit(7), _hit(3)]}}
+
+    qv = np.zeros((1, 8), dtype=np.float32)
+    out = be.search(name="idx", query_vectors=qv, top_k=2)
+
+    assert out == [[7, 3]]
+    be._client.search.assert_called_once()
+    be._client.msearch.assert_not_called()
+
+
+def test_vdb4_multi_query_uses_single_msearch_call():
+    be = _make_backend()
+    # Two query vectors -> two response entries, preserving order.
+    be._client.msearch.return_value = {
+        "responses": [
+            {"hits": {"hits": [_hit(1), _hit(2)]}},
+            {"hits": {"hits": [_hit(9)]}},
+        ]
+    }
+
+    qv = np.zeros((2, 8), dtype=np.float32)
+    out = be.search(name="idx", query_vectors=qv, top_k=2)
+
+    assert out == [[1, 2], [9]]
+    # Exactly one HTTP round-trip for the whole batch, not one per vector.
+    be._client.msearch.assert_called_once()
+    be._client.search.assert_not_called()
+
+    # Body is header+body interleaved: 2 vectors -> 4 entries.
+    body = be._client.msearch.call_args.kwargs.get("body")
+    if body is None and be._client.msearch.call_args.args:
+        body = be._client.msearch.call_args.args[0]
+    assert len(body) == 4
+    assert body[0] == {"index": "idx"}
+    assert "knn" in body[1]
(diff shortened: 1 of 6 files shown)


idevasena:
Thank you @FileSystemGuy! I am still testing each of fixes for 4 bugs mentioned. Here are some initial checks:

1. VDB-1 and VDB-2 are clear correctness wins. The fabricated-percentile bug (identical latency per query in a batch) and the permanently-dead flat_index rebuild guard both check out exactly as described. VDB-1 correctly adds `interval_query_count` — without it, moving to one-entry-per-batch would have broken interval QPS, so this is the right call. Commits are atomic (one file each), well-described, and cumulative QPS (total_queries / wall_elapsed) is untouched and stays correct.
2. VDB-3 seems to be incomplete. Removing the implicit flush() from row_count() is correct, but no callers were updated to `call flush_collection()`. `row_count()` is reached via `base.get_collection_stats()`. Checking the load path 
to confirm none silently read a stale `num_entities` now.

idevasena:
@FileSystemGuy Verified each fix  — all four issues are correctly addressed. Two changes I pushed and added regression tests for your PR before merge:
1. VDB-3 is inconsistent. Audited the callers: the only path reaching row_count() for its count is the base get_collection_stats(), which Milvus/ES/pgvector all override — so no load path silently reads a stale count. But Milvus's get_collection_info() and get_collection_stats() still call col.flush() directly, on the same monitoring paths VDB-3 is trying to keep side-effect-free. So, dropped those two implicit flushes as well so the whole introspection surface is consistent, leaving flush_collection() as the one explicit flush. 
2. Updates regression tests. search_runner and orchestrator have no coverage, so the test plan is all manual. Added three (no live cluster needed): VDB-1 asserts P99 > P50 when batch wall-times vary; VDB-2 asserts flat_index rebuilds truth and never loads the .npz (+ a precomputed-mode guard); VDB-4 covers both the len==1 fast path and the len>1 single-_msearch path. The VDB-4 module is kept separate so it doesn't transitively import pymilvus.

idevasena (approved):

FileSystemGuy:
Please approve:  @russfellows

russfellows (approved):
Curtis says it is good to go, so approving.

PR #402, opened 2026-06-02
storage_tokens_processed undercounts by tensor_parallel factor

FileSystemGuy:
num_tokens = entry_size / self.model_config.kv_cache_size_per_token — entry_size is the TP-sharded per-rank size (1/tensor_parallel of the full entry), but kv_cache_size_per_token is the full unsharded value. Dividing a sharded byte count by an unsharded per-token size undercounts storage_tokens_processed by a factor of tensor_parallel for every NVMe read when tensor_parallel > 1.


idevasena (dismissed):

FileSystemGuy:
Please review: @russfellows

dslik:
Waiting on feedback from @hazemawadalla

hazemawadalla:
I think this  specific pr makes sense logically but actually introduces a bug. correct fix  is 
if self.model_config.kv_cache_size_per_token > 0:
    sharded_bytes_per_token = self.model_config.kv_cache_size_per_token / max(1, self.tensor_parallel)
    num_tokens = entry_size / sharded_bytes_per_token
    self.stats['storage_tokens_processed'] += num_tokens

But I think the way Tensor Parallism  and num-gpus is emulated now needs a full deep review

hazemawadalla (dismissed):
if self.model_config.kv_cache_size_per_token > 0:
    sharded_bytes_per_token = self.model_config.kv_cache_size_per_token / max(1, self.tensor_parallel)
    num_tokens = entry_size / sharded_bytes_per_token
    self.stats['storage_tokens_processed'] += num_tokens

dslik:
@FileSystemGuy Let's make Hazem's suggested change.

FileSystemGuy:
Applied the requested change to `MultiTierCache.access_cache` (`kv_cache_benchmark/kv_cache/cache.py`):

```python
if self.model_config.kv_cache_size_per_token > 0:
    sharded_bytes_per_token = self.model_config.kv_cache_size_per_token / max(1, self.tensor_parallel)
    num_tokens = entry_size / sharded_bytes_per_token
    self.stats['storage_tokens_processed'] += num_tokens
```

Note that the prior version referenced an undefined `kv_cache_size_per_token` (missing the `self.model_config.` prefix), so this also fixes a latent `NameError` on the NVMe-read path.

### Alignment check on the other two `storage_tokens_processed` sites

- **Write path** (`cache.py:733`, `_allocate_cache_inner` -> `nvme` branch): adds the input `num_tokens` directly. Since `size_bytes = kv_cache_size_per_token * num_tokens // tensor_parallel`, the reviewer's `entry_size / sharded_bytes_per_token` formula recovers the same value; using the input is more precise (no rounding). Left as-is.
- **Demote path** (`cache.py:416-422`, `_demote_entry` cpu->nvme): already divides `kv_cache_size_per_token` by `tensor_parallel`, so the TP undercount bug isn't present here. The only stylistic delta vs. the reviewer's snippet is `//` (integer) vs. `/` (float) — same result within ±1 token per eviction. Left as-is to keep the diff minimal; happy to convert it for full consistency if you'd like.

### Regression test

Added `TestStorageTokensProcessedTensorParallel` in `kv_cache_benchmark/tests/test_kv_cache.py`, parametrized over `tensor_parallel in {1, 2, 4}`:

- `test_nvme_write_adds_logical_token_count` — pins the write path at the logical token count.
- `test_nvme_read_adds_logical_token_count` — pins the read path you just fixed. Without the fix this would have raised `NameError`, and a naive `entry_size / kv_cache_size_per_token` would have undercounted by `tensor_parallel`.

### Local test run

```
239 passed, 1 failed, 1 warning in 157.66s
```

The one failure (`TestThreeTierEvictionCascade::test_tier_order_includes_fake_gpu`) is unrelated — it assumes `torch`/CUDA is not installed in the test environment, but `uv sync --extra full` pulls in torch, which causes the GPU backend to initialize. The assertion `'gpu' not in tier_order_before` fails as a result. Same failure occurs on a clean checkout; not caused by this change.

dslik (approved):

PR #404, opened 2026-06-04
Yaml schema validator moves to Pydantic

FileSystemGuy:
Replace Yamale schema with Pydantic v2 validator for system description

Summary

- Replaces the Yamale-based schema.yaml with a Pydantic v2 Python validator (system_description/schema_validator.py) that provides both structural validation (phase 1) and discriminator-based cross-field validation (phase 2)
- Adds pydantic>=2.0 to project dependencies
- Fixes several factual errors in the example YAML files that the new validator caught

What the validator enforces

Structural (phase 1) — field types, required fields, enum membership, numeric minimums, list minimum lengths; equivalent to the previous Yamale schema.

Cross-field / discriminator (phase 2):
- deployment: onprem requires chassis.rack_units, chassis.power, and total_rack_units to be present on all product nodes and clients; also requires rack_units and power on all switches when product_switches is non-empty
- deployment: cloud forbids product_switches, rack_power_supplies, and any chassis.rack_units or chassis.power anywhere in the document
- total_rack_units must equal the sum of product_nodes[i].chassis.rack_units × quantity +
  product_switches[i].rack_units × unit_count (clients are excluded from this sum per spec)
- power_device.min_psus_active must not exceed the total installed PSU count across all psus_configured entries
- remap_time_in_seconds is 0 if and only if both simultaneous_write and simultaneous_read are true
- client_footprint and client_installation must both be 'n/a' or neither
- storage_location: remote or remote_and_local requires networking to be present and non-empty on all product nodes and clients
- storage_location: local forbids product_switches

Testing

101 unit tests in system_description/test_schema_validator.py covering every validation rule with both passing and failing cases. Run with:

uv run pytest system_description/test_schema_validator.py -v


dslik:
Recommend we drop power supply vendor and model, since this information is often not easily available to submitters.

I also recommend that efficiency have an "unrated" option, since not all power supplies have an official rating.

I will submit a PR with more changes once this one is merged.

mlpstorage_py/system_description/schema.yaml
```diff
@@ -0,0 +1,145 @@
...
+---
+#
+# Each individual power supply needs to be described
+#
+power_supply:
+    unit_count:                 int( min=1 )        # Number of power supplies that look like this
+    vendor_name:                str( min=1 )        # The name of the vendor who supplies this power supply
+    model_name:                 str( min=1 )        # The vendor's model name for this power supply
+    power_capacity_watts:       int( min=1 )        # The total output capacity of the power supply
+    efficiency:                 enum( 'Gold', 'Platinum', 'Titanium', 'Ruby' )
```
dslik, line 144: The "efficiency" enum should use the strings defined in the 80 Plus testing standard: https://www.clearesult.com/80plus/80plus-psu-ratings-explained

We should also have an "Unrated" option.

We should also add an "inlet_voltage" item.

russfellows (commented):
Here is my agents answer on this PR:
----
Thank you for pasting the raw content directly—that gives me exactly what we need to do a thorough code review.

Looking over this patch set from Curtis (`FileSystemGuy`) regarding the new **MLPerf Storage / System Description Power Reporting Guidelines**, the intent makes perfect sense: capturing structural data-center nameplate capacity to contrast with active, measured benchmark power.

However, looking strictly at the implementation across these patches, there are several **problematic design inconsistencies, validation issues, and structural bottlenecks** that you should call out in your PR review.

---

## 1. Critical Inconsistencies in Naming & Casing

The most glaring issue in Patch 03 is the sudden shifting of key casing. This will break parsers and frustrate developers writing automated tools to read these submissions.

* **Mixed Casing:** Patch 02 establishes lowercase snake_case anchors (`system_description`, `power_requirements`, `client_nodes`). But Patch 03 suddenly switches to capitalized PascalCase keys for the top-level section:
```yaml
Power:
    Provisioned:            include('power_provisioned',required=True)
    Consumed:               include('power_consumed',required=False)

```


* **Spaces in YAML Keys:** In Patch 03, the schema introduces keys with spaces: `Min PSUs Needed`. While technically legal in YAML, **spaces in keys are highly discouraged** in automation pipelines. It forces downstream processing scripts (like Python or Rust parser engines) to access properties using clunky bracket notation `config["Min PSUs Needed"]` instead of standard object dot-notation `config.min_psus_needed`.

**Recommendation:** Enforce a strict, repository-wide casing standard (ideally standard lowercase `snake_case`) for all keys and variables (e.g., `min_psus_needed`, `power_capacity_watts`).

---

## 2. Brittle Hardware Assumptions vs. Dynamic Arrays

In Patch 02, the schema tries to capture multi-PSU configurations by explicitly hardcoding individual fields up to 6 power supplies:

```yaml
psu1_nameplate_power:   int(min=1,required=True)
psu2_nameplate_power:   int(min=1,required=False)
...
psu6_nameplate_power:   int(min=1,required=False)

```

* **The Problem:** This is deeply brittle. Multi-node multi-chassis architectures, high-density AI server frames, or large enterprise storage arrays frequently deploy more than 6 power distribution blocks or PSUs.
* **The Fix Attempt:** Patch 03 attempts to fix this by pivoting to a block list: `PSUs: list(include('power_supply'))`. This is much better, but Patch 02's `power_detail` block with the hardcoded `psu1_...psu6_` fields is still left in limbo or improperly abstracted.

**Recommendation:** Reject any hardcoded sequential variables (`psu1`, `psu2`). Ensure all power supply tracking is handled exclusively via clean, open-ended sequence lists (`list(...)`).

---

## 3. Treating Booleans as Strings

In Patch 02's schema definitions, boolean fields are declared as string enumerations:

```yaml
shared_capabilities:
    multi_host_support:         enum('True','False')
    simultaneous_write_support: enum('True','False')

```

* **The Problem:** YAML natively supports boolean types (`true` / `false` without quotes). By forcing submitters to input them as literal case-sensitive strings (`'True'`), you introduce silent validation failures. If a submitter writes `multi_host_support: true` natively, a strict parser checking against `enum('True', 'False')` will throw a type error.

**Recommendation:** Define these variables natively as boolean types rather than forcing string enums.

---

## 4. Redundant/Conflicting Power Metrics

The schema asks for both:

1. Individual PSU capacity ratings.
2. An overall computed `design_power` metric.

If a submitter inputs three 1200W PSUs, but types `2500` for `design_power`, the file becomes internally contradictory. The schema currently provides no validation logic to confirm that `design_power` equals the sum or logical aggregate of the specified active power supplies.

**Recommendation:** Clarify in the schema validation framework whether `design_power` is intended to be a programmatic calculation derived from the list, or if it represents a distinct data point (e.g., the thermal envelope limit of the chassis independent of the PSUs).

---

## 5. Structural Interruption

The patch cut off right as Patch 03 was defining the properties for `power_device`. The closing list anchor is cut off mid-character: `PSUs: list(include('power_supply'),`. Make sure that in the final file, the nested `power_supply` definition block is fully realized, validated, and mapping accurately to the DMTF RedFish standard it claims to emulate.

How strict is your working group's tooling regarding backward compatibility for older submission YAMLs if we enforce these naming conventions?

dslik:
> ```yaml
> psu1_nameplate_power:   int(min=1,required=True)
> psu2_nameplate_power:   int(min=1,required=False)
> ...
> psu6_nameplate_power:   int(min=1,required=False)
> ```

I can't find where these "psu#" entries are defined in the updated PR files. PSUs should be a variable list - I think this is already resolved.

> ## 4. Redundant/Conflicting Power Metrics
> The schema asks for both:
> 
> 1. Individual PSU capacity ratings.
> 2. An overall computed `design_power` metric.

We can calculate the design power (sum of all power supply rated wattage minus the sum of all of the redundent power supply rated wattage). Perhaps this would be less error-prone?

mlpstorage_py/system_description/schema.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+---
+#
+# Each individual power supply needs to be described
+#
+power_supply:
+    unit_count:                 int( min=1 )        # Number of power supplies that look like this
+    vendor_name:                str( min=1 )        # The name of the vendor who supplies this power supply
+    model_name:                 str( min=1 )        # The vendor's model name for this power supply
+    power_capacity_watts:       int( min=1 )        # The total output capacity of the power supply
```
dslik, line 143: This should be "nameplate_power_watts", # Regulatory rating for maximum input power draw

FileSystemGuy:
@dslik I made the changes you requested.
@russfellows Unfortunately, you asked your agent to review a much earlier version of the schema.  The issues it identified had already been fixed in this PR.

I believe this PR is ready for your (next) review.

FileSystemGuy:
I also changed the DLIO repo from Russ' repo to the MLCommons one and updated the uv.lock file.

FileSystemGuy:
Resolves issue #406

russfellows (approved):

dslik:
Good to merge.

dslik (approved):

PR #414, opened 2026-06-08
YAML Schema - Update power values

dslik:
As mentioned in #404, this updates the validation schema to drop power supply make and model, and clarifies that we are collecting nameplate (input) power rather than output power.

### DIFF
--- a/mlpstorage_py/system_description/example_NAS.yaml
+++ b/mlpstorage_py/system_description/example_NAS.yaml
@@ -27,9 +27,7 @@ system_under_test:
              power:
                 min_psus_active: 1
                 psus_configured:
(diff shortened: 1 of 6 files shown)


FileSystemGuy:
Could you please change the example_* files to match?  They are the basis for submitters to build off of and we'd like them to pass the schema checker.

dslik:
> Could you please change the example_* files to match? They are the basis for submitters to build off of and we'd like them to pass the schema checker.

Updates to example schemas applied.

FileSystemGuy (approved):

FileSystemGuy:
Please review: @idevasena

idevasena (approved):

PR #420, opened 2026-06-09
fix(kvcache): count per-tier bytes when data reaches a tier via eviction

### LINKED ISSUE #408: KVCache: tier_storage_kv_bytes_written_gb and tier_cpu_kv_bytes_written_gb always 0 when data reaches those tiers via eviction
## Summary

When the KV cache benchmark evicts data from a higher tier (GPU/CPU) into a lower tier (CPU/NVMe), the per-tier byte counters for the destination tier are never incremented. As a result, the reported `tier_cpu_kv_bytes_written_gb` and `tier_storage_kv_bytes_written_gb` (and the bandwidth values derived from them) are `0` in any workload where data primarily reaches those tiers via eviction rather than direct allocation.

## Reproduction

Run any standard config that fills GPU first (the typical multi-tier setup):

```
python kv-cache.py --model llama3.1-8b --duration 600 \
  --num-users 96 --num-gpus 2 --gpu-mem-gb 16 --cpu-mem-gb 32 \
  --cache-dir /path/to/cache --storage-capacity-gb 200 \
  --generation-mode none
```

In `results.json → summary.cache_stats`:

```json
"tier_gpu_kv_bytes_written_gb":      295.15,
"tier_cpu_kv_bytes_written_gb":        0.0,        // actually receiving data
"tier_storage_kv_bytes_written_gb":    0.0,        // actually receiving data
"offloads_cpu":                          0,
"offloads_storage":                    810,        // proves storage IS being written
"tier_storage_write_bandwidth_gbps":   0.0         // derived from the broken counter
```

External verification with `darshan-runtime` (STDIO module) on the NVMe-backed `--cache-dir` shows ~175 GiB written in the same 600 s window.

## Root cause

In `kv_cache/cache.py::_demote_entry()`, the `to_tier in {'cpu','nvme'}` branches increment `offloads_*` and `storage_tokens_processed` but skip the per-tier byte counter:

```python
with self.stats_lock:
    self.stats['evictions'] += 1
    if to_tier == 'cpu':
        self.stats['offloads_cpu'] += 1
        # MISSING: self.stats['tier_cpu_kv_bytes_written'] += size
    elif to_tier == 'nvme':
        self.stats['offloads_storage'] += 1
        # MISSING: self.stats['tier_storage_kv_bytes_written'] += size
        bytes_per_token = ...
```

The other write path (`_admit_entry`, direct allocation) tracks bytes correctly. The unit test in `tests/test_kv_cache.py` asserts `tier_storage_kv_bytes_written_gb > 0`; it passes because it exercises the direct-allocation path, not eviction.

Similarly, the read side of `_demote_entry` (data is read out of `from_tier` to be migrated) does not increment `tier_{from_tier}_kv_bytes_read`.

## Suggested fix

```diff
              try:
                  data, read_timing = self.backends[from_tier].read(key)
                  write_timing = self.backends[to_tier].write(key, data)
                  self.backends[from_tier].delete(key)
+                 with self.stats_lock:
+                     tier_read = 'storage' if from_tier == 'nvme' else from_tier
+                     self.stats[f'tier_{tier_read}_kv_bytes_read'] += size

                  ...

                  with self.stats_lock:
                      self.stats['evictions'] += 1
                      if to_tier == 'cpu':
                          self.stats['offloads_cpu'] += 1
+                         self.stats['tier_cpu_kv_bytes_written'] += size
                      elif to_tier == 'nvme':
                          self.stats['offloads_storage'] += 1
+                         self.stats['tier_storage_kv_bytes_written'] += size
                          bytes_per_token = ...
```

## Verification

Applied the patch locally and ran a 180s reproducer:

| Source | Storage writes (conv_user .npy) |
| --- | --- |
| Darshan (ground truth, STDIO+POSIX) | **129.02 GiB** |
| Bench `tier_storage_kv_bytes_written_gb` (after patch) | **127.46 GiB** |

Delta is ~1.2%, attributable to darshan including the small .npy file header bytes that the bench size accounting excludes.

## Impact

`tier_storage_write_bandwidth_gbps` is the natural metric to report storage bandwidth from `results.json`. Today it returns `0.0` on any default multi-tier setup, forcing users to instrument externally (darshan, perf counters, etc.) to recover the real number.

## Environment

- Bench: `pyproject.toml` version `"3.0.0"` (current `main`)
- Python 3.12, torch 2.6.0+cu118
- 2 × NVIDIA RTX 8000 (48 GB), TP=1
- BeeGFS-backed `--cache-dir`

jay-tau:
Fixes #408

## Summary

`MultiTierCache._demote_entry()` (`kv_cache_benchmark/kv_cache/cache.py`) increments the offload counters and `storage_tokens_processed` when it evicts an entry into a lower tier, but it never increments the **destination tier's byte counter**. As a result, `tier_cpu_kv_bytes_written` / `tier_storage_kv_bytes_written` — and the `*_gb` and `*_write_bandwidth_gbps` values derived from them in `results.json → summary.cache_stats` — report `0` on any multi-tier setup where data reaches CPU or NVMe through **eviction** rather than direct allocation (the typical GPU-first configuration). The read side of the demotion (data read out of the source tier) was likewise untracked.

This is the bug reported in #408, where `offloads_storage` was `810` (data demonstrably written to storage) yet `tier_storage_kv_bytes_written_gb` and `tier_storage_write_bandwidth_gbps` were `0.0`.

## Fix

Inside the existing `stats_lock` block in `_demote_entry()`:
- increment the source tier's `tier_*_kv_bytes_read` counter (the eviction reads the entry out of `from_tier`), and
- increment the destination tier's `tier_*_kv_bytes_written` counter for the `cpu` / `nvme` branches.

This mirrors the accounting already performed on the direct-allocation path (`_allocate_cache_inner`) and the read path (`access_cache`), reusing the same `nvme → storage` stat-name mapping.

## Test

Strengthens `TestThreeTierEvictionCascade::test_full_cascade_gpu_to_cpu_to_nvme_to_delete`, which already forces a GPU→CPU→NVMe cascade (every new entry is admitted to GPU, so CPU/NVMe bytes can only arrive via demotion). New assertions verify:
- the demote-only byte counters (`tier_cpu_kv_bytes_written`, `tier_storage_kv_bytes_written`, `tier_gpu_kv_bytes_read`, `tier_cpu_kv_bytes_read`) are non-zero;
- the user-facing summary exposes non-zero `tier_{cpu,storage}_kv_bytes_written_gb` and `tier_storage_write_bandwidth_gbps`;
- byte conservation — bytes read out of a tier equal the bytes written into the tier below it — pinning the accounting to the demotion path specifically.

This test **fails before** the fix (counters `0` while `offloads_cpu`/`offloads_storage` are non-zero) and **passes after**.

## Verification

- Full KV cache suite: `206 passed, 28 skipped` (skips are CUDA-only).
- The before/after counters were also cross-checked against the `darshan`-measured ground truth in #408 (~1% delta, attributable to `.npy` header bytes excluded from bench size accounting).
- No new `ruff` findings on the changed lines.

### DIFF
--- a/kv_cache_benchmark/kv_cache/cache.py
+++ b/kv_cache_benchmark/kv_cache/cache.py
@@ -409,10 +409,20 @@ class MultiTierCache:
 
                 with self.stats_lock:
                     self.stats['evictions'] += 1
+
+                    # An eviction reads the entry out of from_tier and writes
+                    # it into to_tier.  Track both so the per-tier byte counters
+                    # (and the bandwidth derived from them) reflect data that
+                    # reaches a tier via eviction, not just direct allocation.
+                    from_stats_name = 'storage' if from_tier == 'nvme' else from_tier
+                    self.stats[f'tier_{from_stats_name}_kv_bytes_read'] += size
+
                     if to_tier == 'cpu':
                         self.stats['offloads_cpu'] += 1
+                        self.stats['tier_cpu_kv_bytes_written'] += size
                     elif to_tier == 'nvme':
                         self.stats['offloads_storage'] += 1
+                        self.stats['tier_storage_kv_bytes_written'] += size
                         bytes_per_token = (self.model_config.kv_cache_size_per_token
                                           // max(1, self.tensor_parallel))
                         if bytes_per_token > 0:
--- a/kv_cache_benchmark/tests/test_kv_cache.py
+++ b/kv_cache_benchmark/tests/test_kv_cache.py
@@ -2492,6 +2492,43 @@ class TestThreeTierEvictionCascade:
         assert cache.stats['offloads_storage'] > 0, \
             "At least one CPU → NVMe demotion should have occurred"
 
+        # --- Per-tier byte counters must reflect eviction-driven I/O ---
+        # New entries are always admitted to GPU (the top tier); data only
+        # reaches CPU and NVMe here via _demote_entry().  Regression guard for
+        # issue #408: those demote paths must increment the destination tier's
+        # written-bytes counter (and the source tier's read-bytes counter),
+        # otherwise tier_{cpu,storage}_kv_bytes_written_gb report 0 even though
+        # offloads_{cpu,storage} prove data is moving.
+        assert cache.stats['tier_cpu_kv_bytes_written'] > 0, \
+            "GPU → CPU demotions must increment tier_cpu_kv_bytes_written"
+        assert cache.stats['tier_storage_kv_bytes_written'] > 0, \
+            "CPU → NVMe demotions must increment tier_storage_kv_bytes_written"
+        assert cache.stats['tier_gpu_kv_bytes_read'] > 0, \
+            "GPU → CPU demotions read out of GPU; tier_gpu_kv_bytes_read must rise"
+        assert cache.stats['tier_cpu_kv_bytes_read'] > 0, \
+            "CPU → NVMe demotions read out of CPU; tier_cpu_kv_bytes_read must rise"
+
+        # Byte conservation: this test only allocates (no access_cache reads),
+        # and every admit lands on GPU, so the only reads out of GPU/CPU are
+        # demotions and the only writes into CPU/NVMe are demotions.  Each
+        # demote moves exactly `size` bytes from from_tier to to_tier, so the
+        # read-out total of a tier must equal the written-in total of the tier
+        # below it.  This pins the accounting to demotion specifically.
+        assert cache.stats['tier_gpu_kv_bytes_read'] == cache.stats['tier_cpu_kv_bytes_written'], \
+            "Every byte read out of GPU during demotion must be written into CPU"
+        assert cache.stats['tier_cpu_kv_bytes_read'] == cache.stats['tier_storage_kv_bytes_written'], \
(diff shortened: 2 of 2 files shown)


dslik:
@hazemawadalla , can you review?

jay-tau:
recheck

hazemawadalla:
thanks let me check it one more time today

dslik:
> thanks let me check it one more time today

@hazemawadalla If this change looks good to you, can you open a new PR so that we can get this in? Resolving the CLA issue may take some time.

jay-tau:
Do I need to do anything beyond replying to the mail that I
received related to the CLA?

Joel

hazemawadalla:
approved looks good

dslik:
> Do I need to do anything beyond replying to the mail that I received related to the CLA? Joel

You will need to sign the CLA, then send it back to MLCommons. That will unblock the cla-check.

jay-tau:
recheck

jay-tau:
I've already signed and submitted the CLA. Hopefully, it picks it up

FileSystemGuy:
Yay!  The CLA check has passed, we just need @dslik to do a final review then merge.

dslik (approved):

PR #424, opened 2026-06-09
Fixes to multi-node VectorDB benchmark orchestration: Align with mlpstorage's MPI-Native 

idevasena:
This PR fixes multi-node initiation and aggregation support for the VectorDB benchmark through `mlpstorage vectordb`, covering:
```
vectordb datagen / load
vectordb run --mode timed
vectordb run --mode query_count
vectordb run --mode sweep
```

MPI ranks use mpi4py collectives for coordination and metric aggregation. Rank-local detailed outputs are written under a node-local `--rank-output-dir`, and the final aggregated summary is emitted to the launcher through MPI-gathered metrics and stdout JSON.

This addresses failures seen when client nodes can all reach the same Milvus/vector DB endpoint but do not share a writable filesystem. The previous implementation passed a shared `--base-output-dir` `/shared/vdb_results/...` to `vdb-mpi-wrapper`, then attempted vdb-aggregate against that same path; rank 1 failed when it could not create/access /shared/vdb_results, causing the distributed load to abort.

#### Key behavior

```
--coordination mpi
  passes --coordination mpi, --rank-output-dir, and --run-id to vdb-mpi-wrapper
  skips shared-filesystem vdb-aggregate
  parses VDB_MULTI_NODE_SUMMARY_JSON emitted by the MPI wrapper
  writes launcher-local summary files under --results-dir
```


FileSystemGuy (approved):
Please review: @dslik @bbelgodere

PR #441, opened 2026-06-15
accumulation: lock down current behavior and harden discovery/reservation

FileSystemGuy:
## Summary

Foundation work for incremental results accumulation, in two atomic commits:

- **PR 1 (test scaffolding)** — `tests/unit/test_accumulation.py` synthesizes per-benchmark results trees under `tmp_path` and drives the **real** `get_runs_files()` walk (no mocking of discovery). Covers 17 scenarios the existing mocked tests in `test_reporting.py` couldn't: heterogeneous trees (training + checkpointing), the N=5 gate in `TrainingSubmissionRulesChecker`, multi-workload separation (5x unet3d + 5x resnet50), and negative cases (corrupt JSON, missing manifests, multiple metadata files, nonexistent dirs, symlinks).

- **PR 2 (fixes for what PR 1 surfaced)** —
  - `DLIOResultParser.parse()` now raises instead of silently producing `benchmark_type=None`.
  - `MultiRunRulesChecker.check_benchmark_type_consistency` defends against callers bypassing `BenchmarkVerifier`'s dispatch-time type guard.
  - `get_runs_files` uses `followlinks=True` so symlinked runs accumulate cleanly.
  - New `mlpstorage_py/run_directory.py` provides exclusive-create + bounded one-second forward bumps on sub-second timestamp collisions, replacing the prior `os.makedirs(exist_ok=True)` in `Benchmark.__init__`. Tests in `test_benchmarks_kvcache.py` / `test_benchmarks_vectordb.py` that pre-created the output dir under the old semantics are updated.

The bug-surfacing tests from PR 1 are flipped to assert the fixed contracts in PR 2.

This PR is the prerequisite for two follow-up PRs (each code-owner-scoped) targeting this branch:
- `feat(vectordb): add --vdb-engine and include engine in results-dir path` (VDB)
- `feat(kvcache): include model in results-dir path` (KVCache)

## Test plan

- [ ] `pytest tests/unit -q` — full suite passes (1249 pass / 4 skip / 1 pre-existing failure unrelated to this work)
- [ ] `pytest tests/unit/test_accumulation.py -v` — all 25 accumulation tests green
- [ ] No production code change requires manual smoke; behavior verified through tests

### DIFF
--- a/mlpstorage_py/benchmarks/base.py
+++ b/mlpstorage_py/benchmarks/base.py
@@ -48,6 +48,10 @@ from functools import wraps
 from pyarrow.ipc import open_stream
 
 from mlpstorage_py.config import PARAM_VALIDATION, DATETIME_STR, MLPS_DEBUG, EXEC_TYPE
+from mlpstorage_py.run_directory import (
+    DEFAULT_COLLISION_BUMP_BUDGET,
+    reserve_run_directory,
+)
 from mlpstorage_py.debug import debug_tryer_wrapper
 from mlpstorage_py.interfaces import BenchmarkInterface, BenchmarkConfig, BenchmarkCommand
 from mlpstorage_py.mlps_logging import setup_logging, apply_logging_options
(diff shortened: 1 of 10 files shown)


mlpstorage_py/benchmarks/base.py
```diff
@@ -802,6 +803,23 @@ def generate_output_location(self) -> str:
             raise ValueError('No benchmark specified. Unable to generate output location')
         return generate_output_location(self, self.run_datetime)
 
+    _COLLISION_BUMP_BUDGET = DEFAULT_COLLISION_BUMP_BUDGET
+
+    def _reserve_run_directory(self) -> str:
+        """Atomically reserve a unique run directory, updating run_datetime
+        if a collision pushes the timestamp forward. See
+        mlpstorage_py.benchmarks.run_directory.reserve_run_directory.
```
idevasena, line 811: Minor:
`run_directory.py` lives at `mlpstorage_py/run_directory.py`, but `_reserve_run_directory's` docstring and the test section header both reference `mlpstorage_py/benchmarks/run_directory.py`.

idevasena (approved):

russfellows (approved):
Approving per Curtis.

PR #442, opened 2026-06-15
vectordb: add --vdb-engine and include engine in results-dir path

FileSystemGuy:
## Summary

Adds an engine component to the VectorDB results-dir path so multiple engines (today: milvus; future: elasticsearch, etc.) accumulate without collision.

**Path change:** \`<results_dir>/vector_database/<engine>/<command>/<datetime>/\` (was \`vector_database/<command>/<datetime>/\`).

Changes (all VDB-scoped):
- \`config.py\`: \`VDB_ENGINES = [\"milvus\"]\`, \`VDB_ENGINE_DEFAULT = \"milvus\"\`.
- \`cli/vectordb_args.py\`: \`--vdb-engine\` on all subcommands (datasize/datagen/run).
- \`rules/utils.py\`: \`generate_output_location\` for \`vector_database\` inserts the engine; raises if missing.
- \`benchmarks/vectordbbench.py\`:
  - \`__init__\` mirrors \`args.vdb_engine\` into \`args.model\` before \`super().__init__()\`, so the existing metadata extractor and workload grouping (keyed on \`(model, accelerator)\`) treat distinct engines as distinct workloads.
  - **Removes \`\"model\": self.config_name\` from the metadata override.** That line had clobbered the engine and would have merged distinct engines sharing a config file into one workload bucket. \`config_name\` is still preserved as \`vectordb_config\`.
- \`reporting/directory_validator.py\`: documented expected structure updated.
- \`tests/unit/test_accumulation.py\`: \`make_vectordb_run\` takes \`engine\`; positive/negative path tests; converts the \"documents limitation\" test from the foundation PR into \`test_vectordb_runs_distinguished_by_engine\`.
- \`tests/unit/test_benchmarks_vectordb.py\`: fixtures gain \`vdb_engine\`; renames \`test_metadata_model_uses_config_name\` → \`test_metadata_model_is_engine_config_preserved_separately\` to assert the new contract.

**Backwards compatible:** old \`vector_database/<command>/<datetime>/\` trees still discover via \`get_runs_files\` (they appear with \`model=None\` and group separately from new engine-tagged runs in reports).

## Merge order

Targets the foundation branch (#441). Merge that first; then this PR's diff against main will be just the VDB changes.

## Test plan

- [ ] \`pytest tests/unit -q\` — full suite passes
- [ ] \`pytest tests/unit/test_accumulation.py tests/unit/test_benchmarks_vectordb.py -v\` — VDB and accumulation suites green

### DIFF
--- a/mlpstorage_py/benchmarks/vectordbbench.py
+++ b/mlpstorage_py/benchmarks/vectordbbench.py
@@ -50,6 +50,14 @@ class VectorDBBenchmark(Benchmark):
     SUMMARY_PREFIX = "VDB_MULTI_NODE_SUMMARY_JSON="
 
     def __init__(self, args, **kwargs):
+        # Mirror the engine into args.model so the existing
+        # metadata extractor (BenchmarkInstanceExtractor) and workload
+        # grouping (ReportGenerator._process_workload_groups, keyed on
+        # (model, accelerator)) treat distinct engines as distinct
+        # workloads — accumulated runs from milvus vs a future engine
+        # stay correctly separated.
+        if hasattr(args, "vdb_engine") and getattr(args, "model", None) is None:
+            args.model = args.vdb_engine
         super().__init__(args, **kwargs)
 
         self.command_method_map = {
@@ -974,7 +982,11 @@ class VectorDBBenchmark(Benchmark):
             {
                 "vectordb_config": self.config_name,
                 "vectordb_config_file": self.config_file,
-                "model": self.config_name,
+                # NB: do NOT set "model" here — the base class records the
+                # engine (via args.model = args.vdb_engine in __init__) and
+                # downstream workload grouping in ReportGenerator keys on it.
+                # Overwriting with config_name would merge distinct engines
+                # that happen to share a config file into one workload.
                 "host": getattr(self.args, "host", "127.0.0.1"),
                 "port": getattr(self.args, "port", 19530),
                 "collection": getattr(self.args, "collection", None),
--- a/mlpstorage_py/cli/vectordb_args.py
+++ b/mlpstorage_py/cli/vectordb_args.py
@@ -26,6 +26,8 @@ from mlpstorage_py.config import (
     VECTOR_DTYPES,
     VECTORDB_DEFAULT_RUNTIME,
     VDB_BENCHMARK_MODES,
+    VDB_ENGINE_DEFAULT,
+    VDB_ENGINES,
     VDB_INDEX_TYPES,
     VDB_INDEX_TYPES_CLOSED,
 )
@@ -205,6 +207,20 @@ def _add_vectordb_core_args(parser, command, index_choices):
     """
     # Set defaults for open-gated attrs so they always exist in the namespace.
     parser.set_defaults(loops=1, params='', allow_invalid_params=False)
+
+    # The engine identifies which vector database implementation is being
+    # benchmarked (milvus today). It is recorded in the results-dir path
+    # and metadata so accumulated runs from different engines coexist.
+    parser.add_argument(
+        '--vdb-engine',
+        choices=VDB_ENGINES,
+        default=VDB_ENGINE_DEFAULT,
+        help=(
+            "Vector database engine being benchmarked. "
+            "Recorded in the result path so multiple engines can accumulate "
+            "in one --results-dir without collision."
+        ),
+    )
 
     # ---- Common args for datagen and run ----
     if command in ("datagen", "run"):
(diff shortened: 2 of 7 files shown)


FileSystemGuy:
@idevasena I'm not sure that you want this or not, so please feel free to suggest changes or even say "no thanks".

mlpstorage_py/reporting/directory_validator.py
```diff
@@ -246,6 +246,12 @@ def get_expected_structure_help(self) -> str:
             checkpointing_llama3-8b_metadata.json
             summary.json
 
+    vector_database/
```
idevasena, line 249: Updating this doc block with below:

```
vector_database/
  milvus/                          # VDB engine (--vdb-engine)
    diskann/                       # index type (--index-type)
      run/
        20250115_160000/
          vector_database_20250115_160000_metadata.json
```

tests/unit/test_accumulation.py
```diff
@@ -154,22 +154,24 @@ def make_checkpointing_run(
 def make_vectordb_run(
```
idevasena, line 154: extend `make_vectordb_run` in `test_accumulation.py` with an `index="diskann"` param and a `vector_database/<engine>/<index>/<command>/<datetime>` path; add `test_vectordb_path_includes_index` and `test_vectordb_path_requires_index` mirroring the engine tests; and add a `test_vectordb_runs_distinguished_by_index` that puts milvus+diskann and milvus+hnsw in one tree and asserts they don't collide.

mlpstorage_py/benchmarks/vectordbbench.py
```diff
@@ -974,7 +982,11 @@ def metadata(self) -> Dict[str, Any]:
             {
                 "vectordb_config": self.config_name,
```
idevasena, line 983: index belongs in metadata too. Keep grouping on engine via model, but record index so reports can break out by it:
```
# in the metadata override dict, alongside vectordb_config:
"vectordb_index": getattr(self.args, "index_type", None),
```

As a follow-up (not in this PR): Optionally make grouping (engine, index)-aware later by mirroring a composite into model (e.g. f"{engine}/{index}") — but that's a bigger report-generator decision.

mlpstorage_py/cli/vectordb_args.py
idevasena, line 200: `--index-type` is currently only on datasize. Promote it to a shared core arg so `datagen/run` also populate `index_type` (otherwise the path build raises for those commands).

In `_add_vectordb_core_args`, add once (using the existing index_choices), and remove the duplicate per-command definitions to avoid argparse conflicts:
```
parser.add_argument(
    '--index-type',
    choices=index_choices,
    default="DISKANN",
    help="Index type; recorded in the results-dir path so DISKANN/HNSW/AISAQ "
         "runs accumulate without collision.",
)
```
May be we can keep a distinct flag name, alias `--vdb-index` with dest='index_type' to keep one source of truth.

mlpstorage_py/rules/utils.py
```diff
@@ -162,7 +162,14 @@ def generate_output_location(benchmark, datetime_str=None, **kwargs) -> str:
         output_location = os.path.join(output_location, datetime_str)
 
     elif benchmark.BENCHMARK_TYPE == BENCHMARK_TYPES.vector_database:
+        engine = getattr(benchmark.args, "vdb_engine", None)
```
idevasena, line 165: insert index after engine:
```
elif benchmark.BENCHMARK_TYPE == BENCHMARK_TYPES.vector_database:
    engine = getattr(benchmark.args, "vdb_engine", None)
    if not engine:
        raise ValueError(
            "VectorDB engine is required for output location "
            "(set --vdb-engine on the CLI)."
        )
    index = getattr(benchmark.args, "index_type", None)
    if not index:
        raise ValueError(
            "VectorDB index type is required for output location "
            "(set --index-type on the CLI)."
        )
    output_location = os.path.join(output_location, benchmark.BENCHMARK_TYPE.name)
    output_location = os.path.join(output_location, engine)
    output_location = os.path.join(output_location, str(index).lower())  # normalize
    output_location = os.path.join(output_location, benchmark.args.command)
    output_location = os.path.join(output_location, datetime_str)
```

idevasena (commented):
Thank you @FileSystemGuy for this PR. As discussed, I suggest extending the path one level deeper to `vector_database/<engine>/<index>/<command>/<datetime>/`. The index type is the biggest driver of both storage footprint and recall/QPS, and CLOSED-vs-OPEN submissions are differentiated primarily by index params. `index_type` already exists as a CLI arg; I included changes to individual files as comments for this additive change.

russfellows (approved):
Approving per Curtis.

PR #443, opened 2026-06-15
kvcache: include model in results-dir path

FileSystemGuy:
## Summary

Adds a model component to the KVCache results-dir path so multiple model variants (llama2-7b, llama3.1-8b, llama3.1-70b-instruct, etc.) accumulate without collision.

**Path change:** \`<results_dir>/kv_cache/<model>/<command>/<datetime>/\` (was \`kv_cache/<command>/<datetime>/\`).

Changes (all kvcache-scoped):
- \`rules/utils.py\`: \`generate_output_location\` for \`kv_cache\` reads \`benchmark.args.model\` and inserts it as a path component; raises if missing.
- \`benchmarks/kvcache.py\`:
  - \`__init__\` defaults \`args.model = KVCACHE_MODEL_DEFAULT\` before \`super().__init__()\`. This is required because the closed-mode CLI does not expose \`--model\` (see \`_add_kvcache_model_arguments\` only being added for open/whatif), and the base class needs \`args.model\` to compute the output path and metadata.
  - Collapses the now-redundant \`getattr(args, 'model', 'llama3.1-8b')\` fallback to a direct \`args.model\` read, and replaces the previously hardcoded string with the constant.
- \`tests/unit/test_accumulation.py\`: \`make_kvcache_run\` takes \`model\`; new path-generation positive/negative tests; \`test_kvcache_runs_distinguished_by_model\`.
- \`tests/unit/test_benchmarks_kvcache.py\`: new \`test_metadata_closed_mode_defaults_model\` exercises the closed-mode default flowing through \`Benchmark.metadata\`.

**Backwards compatible:** existing \`kv_cache/<command>/<datetime>/\` trees still discover via \`get_runs_files\` (they appear with \`model=None\` and group separately from new runs in reports).

## Merge order

Targets the foundation branch (#441). Merge that first; then this PR's diff against main will be just the kvcache changes.

## Test plan

- [ ] \`pytest tests/unit -q\` — full suite passes
- [ ] \`pytest tests/unit/test_accumulation.py tests/unit/test_benchmarks_kvcache.py -v\` — kvcache and accumulation suites green

### DIFF
--- a/mlpstorage_py/benchmarks/kvcache.py
+++ b/mlpstorage_py/benchmarks/kvcache.py
@@ -26,6 +26,7 @@ from mlpstorage_py.benchmarks.base import Benchmark
 from mlpstorage_py.config import (
     BENCHMARK_TYPES,
     KVCACHE_DEFAULT_DURATION,
+    KVCACHE_MODEL_DEFAULT,
 )
 from mlpstorage_py.interfaces import BenchmarkCommand
 from mlpstorage_py.utils import generate_mpi_prefix_cmd, MLPSJsonEncoder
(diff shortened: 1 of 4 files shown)


FileSystemGuy:
@dslik @hazemawadalla I'm not sure you want this, or even if it is appropriate, so feel free to suggest changes or even "no thanks".

dslik (commented):
Since each run of KV cache includes nine runs with two models, this would be better organized by timestamp rather than by model.

FileSystemGuy:
@dslik I think this PR is not needed, please let me know what you think.

The output pathnames in the results-directory already contain a 'datetime' directory name so that multiple runs of the same workload accumulate in the results-directory rather than colliding.  That was my original reason for letting claude do this work, so I'm thinking that maybe I just close this PR and leave the output pathnames the way they are.

dslik:
Let's close this without change.

Issue #444, opened 2026-06-15
kvcache datasize whatif crashes: 'Namespace' object has no attribute 'loops'


FileSystemGuy:
## Summary

\`mlpstorage whatif kvcache datasize\` crashes with an attribute error before the benchmark even starts.

## Reproducer

\`\`\`
$ ./mlpstorage whatif kvcache datasize \\
    --results-dir /tmp/mlps-results \\
    --gpu-mem-gb 16 \\
    --cpu-mem-gb 32
\`\`\`

## Error

\`\`\`
ERROR: Unexpected error: 'Namespace' object has no attribute 'loops'
ERROR: An internal error occurred: 'Namespace' object has no attribute 'loops'
This is likely a bug in MLPerf Storage.
\`\`\`

## Observations

- The crash happens during the run-configuration summary (printed before \"Validating environment\"), so it's in the argparse/dispatch layer, not the benchmark \`__init__\`.
- \`loops\` is registered only on the \`run\` subparser for kvcache (not on \`datasize\`), but something downstream of arg parsing reads \`args.loops\` unconditionally.
- Only verified in \`whatif\` mode; haven't checked \`closed\` or \`open\`.
- Discovered while writing an end-to-end integration test for the accumulate-results work; this command was a candidate smoke test for kvcache's results-dir layout but is blocked by this issue.

## Environment

- Branch: \`main\` (verified independently of accumulate-results work — the failure is pre-existing)
- Python 3.12.3, Open MPI 4.1.6, WSL2 (Ubuntu)

## Suggested fix direction

Either add a \`loops\` default to the kvcache datasize subparser (\`parser.set_defaults(loops=1)\` in \`mlpstorage_py/cli/kvcache_args.py\`) or gate the reader of \`args.loops\` on \`getattr(args, 'loops', None)\`.

PR #446, opened 2026-06-15
accumulation: real MPI end-to-end integration tests (slow)

FileSystemGuy:
## Summary

Adds `tests/integration/test_real_accumulation.py` with **12 `@pytest.mark.slow` integration tests** that drive the real `mlpstorage` CLI via subprocess against the real MPI runtime, verifying that the accumulation pipeline works end-to-end for all three benchmark types.

End-to-end runtime: **101s on WSL2** (against a tiny dataset; the goal is functional verification, not benchmark scoring).

### Training (5 tests, real mpirun)
- Runs `whatif training unet3d datagen file` once and `whatif training unet3d run file` twice into one results-dir.
- Asserts path layout (`<results>/training/unet3d/<command>/<datetime>/`).
- Asserts real metadata has every `BenchmarkRunData` required field plus `executed_command` starting with `mpirun`.
- Asserts `get_runs_files` discovers all 3 runs (1 datagen + 2 run).
- Asserts `TrainingSubmissionRulesChecker.check_num_runs` fires INVALID at N=2 (the same gate the unit tests exercise, now against real output).
- Asserts back-to-back runs land in distinct timestamp directories (regression guard for the exclusive-create + bump path in `mlpstorage_py/run_directory.py`).

### VectorDB (3 tests, real CLI)
- Runs `whatif vectordb datasize` (no Milvus server or pymilvus needed — pure storage calculation).
- Asserts path includes the engine: `vector_database/milvus/datasize/<datetime>/`.
- **Asserts metadata `model == \"milvus\"`, NOT `\"default\"`** — direct regression guard for the metadata-override bug that PR #442 fixes. Without that fix, distinct engines sharing a config file would have merged into one workload bucket.
- Asserts `vectordb_config` is preserved separately.
- Asserts `get_runs_files` attributes the run to `model=milvus`.

### KVCache (3 tests, real CLI)
- Runs `whatif kvcache run --dry-run` (avoids needing llama weights; the reservation + metadata code in `Benchmark.__init__` runs before benchmark execution so path assertions are unaffected).
- Asserts path includes the model: `kv_cache/tiny-1b/run/<datetime>/`.
- Asserts metadata records `model = \"tiny-1b\"`; backward-compat `kvcache_model` field also preserved.
- Asserts `get_runs_files` attributes the run to `model=tiny-1b`.
- Deliberately skips `kvcache datasize whatif`, which crashes with a pre-existing bug unrelated to this work (tracked as #444).

### Heterogeneous tree (1 test)
- Combines all three benchmark types' fixture outputs into a single results-dir via symlinks and verifies `get_runs_files` discovers them with correct per-type attribution. Exercises the `followlinks=True` change in `get_runs_files` (from #441) under real fixture output.

## Guards against flakiness / wrong-environment runs

- `@pytest.mark.slow` plus `addopts = \"-m 'not slow'\"` in `pyproject.toml` — the default suite excludes these. Opt in with `pytest -m slow` (or `pytest -m ''`).
- Skip-if predicate when `MLPS_SKIP_INTEGRATION` env var is set (CI escape hatch), when the `mlpstorage` shim is missing/not executable, or when `mpirun` / `uv` are off PATH.
- `--allow-run-as-root` passed to every CLI call (CI runners often run as root; OpenMPI refuses by default).
- 600s subprocess timeout per CLI call.
- Tiny dataset (`dataset.num_files_train=10`, ~1 GB) keeps total runtime under ~2 minutes. AU% is always 0.0 at this scale; the test is functional, not a benchmark score.

## Merge order

This PR depends on **all three** prior PRs in this effort:
- #441 (Foundation — test scaffolding + discovery/reservation fixes)
- #442 (VectorDB — `--vdb-engine` + path layout)
- #443 (KVCache — model in path layout)

The integration test reaches into production code from each, so the branch carries cherry-picks of those commits (visible in this PR until they merge). Once #441, #442, and #443 merge, this branch can be rebased onto `main` and the diff collapses to just the new integration test file plus the `pyproject.toml` marker registration.

## Test plan

- [ ] `pytest tests/unit -q` — full unit suite still green (1251 pass / 4 skip on this branch)
- [ ] `pytest tests/integration/test_real_accumulation.py -m slow -v` — all 12 integration tests pass
- [ ] `pytest tests/integration/test_real_accumulation.py --collect-only` (no `-m`) — 0 collected (slow tests correctly excluded by default)


russfellows (dismissed):
Approving per Curtis.

russfellows (dismissed):
Approving per Curtis.

FileSystemGuy:
I'm working on resolving the conflicts

russfellows (approved):
Let's roll.

PR #458, opened 2026-06-16
feat(system_description): add drive media_type and performance block

FileSystemGuy:
Add a required `media_type` enum (HDD, TLC, QLC, other) to drive_instance so every submission discloses whether storage is spinning disk or flash, and which NAND cell type for flash.

Add an optional `performance` sub-block carrying vendor-published spec-sheet numbers for an individual drive. The whole block is optional, but if present seq_read_MBps, seq_write_MBps, random_read_IOPS, and random_write_IOPS are all required. form_factor and interface_speed are optional within the block.

Updates schema.yaml, the pydantic validator, all example YAMLs that contain drive entries, and the schema_validator unit tests (16 new tests covering media_type enforcement and the performance-block optionality contract).

### DIFF
--- a/mlpstorage_py/system_description/example_NFS.yaml
+++ b/mlpstorage_py/system_description/example_NFS.yaml
@@ -42,13 +42,29 @@ system_under_test:
               - vendor_name: Micron
                 model_name: X9000 TLC
                 interface: nvme
+                media_type: TLC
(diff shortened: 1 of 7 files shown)


russfellows (approved):
Curtis said approve all.

PR #470, opened 2026-06-18
chore(deps): track DLIO main @ PR #25 merge (fixes #455)

### LINKED ISSUE #455: `dlio_sampler` uses `ceil(N/size)` causing inter-epoch deadlock when `num_samples` is not divisible by `comm_size`
## Summary

When the total number of samples is not evenly divisible by `comm_size`, the `dlio_sampler` class in `torch_data_loader.py` assigns fewer samples to the last rank than to all other ranks. With `drop_last=True` (the default for training), the last rank produces one fewer batch per epoch. The training loop calls `comm.barrier()` once per batch and once at end of epoch, so the mismatched batch count causes ranks to enter incompatible collectives and deadlock at every epoch boundary.

## Root cause

`torch_data_loader.py`, lines 418-423:

```python
class dlio_sampler(Sampler):
    def __init__(self, rank, size, num_samples, epochs):
        samples_per_proc = int(math.ceil(num_samples / size))   # ← ceil
        start_sample = self.rank * samples_per_proc
        end_sample = (self.rank + 1) * samples_per_proc - 1
        if end_sample > num_samples - 1:
            end_sample = num_samples - 1                        # ← clamp last rank
        self.indices = list(range(start_sample, end_sample + 1))
```

`math.ceil(N / size)` overshoots for the last rank, then the clamp at line 422 cuts it back. When `N % size != 0`, the last rank ends up with fewer samples than every other rank.

### Example (N=100, 7 ranks, batch_size=3)

| Ranks | samples_per_proc | Actual samples | Batches (drop_last) |
|-------|----------------:|---------------:|--------------------:|
| 0..5 | 15 | 15 | **5** |
| 6 | 15 | **10** (clamped) | **3** |

The last rank gets 2 fewer batches. The same pattern applies at any scale where `N % size != 0`.

### How the deadlock forms

`main.py` `_train()` has two barriers:

```python
# line 456 — called once per batch step
self.comm.barrier()

# line 478 — called once at end of epoch
self.comm.barrier()
```

Execution trace (using the N=100, 7 ranks, batch=3 example above):

1. All 7 ranks enter the training loop. Ranks 0..5 do 5 batches; rank 6 does 3.
2. For the first 3 steps, all ranks call the line 456 barrier together. No problem.
3. Step 4: ranks 0..5 call barrier at line 456. Rank 6 has finished its loop and calls barrier at line 478 (end of epoch). MPI matches these because it does not distinguish barrier call sites.
4. All ranks unblock. Ranks 0..5 continue stepping; rank 6 has already moved on to `reconfigure()` for the next epoch.
5. Step 5: ranks 0..5 call barrier at line 456. Rank 6 calls `comm.reduce()` from `reconfigure()`. These are incompatible collectives. **Permanent deadlock.**

## Proposed fix

Replace `ceil` with floor division so all ranks get the same sample count:

```python
samples_per_proc = num_samples // size
start_sample = self.rank * samples_per_proc
end_sample = (self.rank + 1) * samples_per_proc - 1
self.indices = list(range(start_sample, end_sample + 1))
```

This drops at most `size - 1` samples per epoch. The clamp is no longer needed since `floor` never overshoots.

## Acceptance criteria

- [ ] All ranks produce the same number of batches per epoch regardless of `N % size`.
- [ ] Multi-epoch runs complete without hanging at epoch boundaries.
- [ ] Dropped sample count is at most `size - 1` per epoch.
- [ ] Unit test: `N=100, size=7, batch=3, drop_last=True` produces equal batch counts on all ranks.

## Impact

- Any `total_samples` not divisible by `comm_size` deadlocks after the first epoch. Since dataset sizes are rarely exact multiples of the rank count, this affects most multi-node multi-epoch runs.
- The failure mode is silent: no error, no timeout, just permanent CPU spin. Operators have to manually cancel and have no diagnostic pointing at the sampler.

FileSystemGuy:
## Summary

Bumps the `dlio-benchmark` git pin to pick up [DLIO PR #25](https://github.com/mlcommons/DLIO_local_changes/pull/25) (`cbe20010`), which fixes the inter-epoch sampler deadlock reported in #455.

- Pin: `1d11f982` → `cbe20010`
- Storage version: `3.0.12` → `3.0.13`
- `uv.lock` regenerated.

## Why

`dlio_sampler` was using `math.ceil(num_samples / comm_size)` with a clamp on the last rank. When `num_samples % comm_size != 0` the last rank produced fewer batches per epoch and the per-step / end-of-epoch `MPI_Barrier`s in `main._train()` matched across iterations — a silent CPU-spinning deadlock at the next epoch boundary with no diagnostic.

PR #25 replaces `ceil + clamp` with floor division at three call sites (`torch_data_loader.dlio_sampler`, `config.build_sample_map_iter`, `config.get_global_map_index`), emits a rank-0 warning when the floor drops samples, and fixes a pre-existing `Sampler.__len__` contract violation (caught by @idevasena during review) so `len(sampler) == len(list(iter(sampler)))`. Includes three regression tests.

Closes #455.

## Test plan

- [ ] `mlpstorage --version` reports `3.0.13`.
- [ ] `pip install -e .` resolves DLIO at `cbe20010`.
- [ ] Multi-rank run with `num_samples % comm_size != 0` (e.g. `unet3d` configured so `num_files_train` is not a multiple of `num-accelerators`) completes past epoch 1 without hanging.
- [ ] DLIO emits the "dropping N sample(s)" warning on rank 0 when N is uneven.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "mlpstorage"
-version = "3.0.12"
+version = "3.0.13"
 description = "MLPerf Storage Benchmark Suite"
 readme = "README.md"
 license = {text = "Apache-2.0"}
@@ -131,7 +131,8 @@ torchaudio = [{ index = "pytorch-cpu" }]
 #   - DLIO PR #22 fix for storage #415 (mpi4py auto-init in spawn workers)
 #   - DLIO PR #23 fix for storage #391 part 2 (sudo -n + warn-once for drop_caches)
 #   - DLIO PR #24 fix for storage #448 (per-node worker-memory budget)
-dlio-benchmark = { git = "https://github.com/mlcommons/DLIO_local_changes.git", rev = "1d11f9820b71db1d59b41283bc8c57db0c42cb7f" }
+#   - DLIO PR #25 fix for storage #455 (sampler floor division — equal per-rank shards)
+dlio-benchmark = { git = "https://github.com/mlcommons/DLIO_local_changes.git", rev = "cbe200103b05fb65f04110157a7d0dd6d6d88ead" }
 
 [dependency-groups]
 dev = [
--- a/uv.lock
+++ b/uv.lock
@@ -237,7 +237,7 @@ wheels = [
 [[package]]
 name = "dlio-benchmark"
 version = "3.0.2"
-source = { git = "https://github.com/mlcommons/DLIO_local_changes.git?rev=814f3ff400c865a294bfc70714a3289efbfa83ac#814f3ff400c865a294bfc70714a3289efbfa83ac" }
+source = { git = "https://github.com/mlcommons/DLIO_local_changes.git?rev=cbe200103b05fb65f04110157a7d0dd6d6d88ead#cbe200103b05fb65f04110157a7d0dd6d6d88ead" }
 dependencies = [
     { name = "dgen-py", marker = "sys_platform == 'linux'" },
     { name = "h5py", marker = "sys_platform == 'linux'" },
@@ -518,7 +518,7 @@ wheels = [
 
 [[package]]
 name = "mlpstorage"
-version = "3.0.12"
+version = "3.0.13"
 source = { editable = "." }
 dependencies = [
     { name = "dlio-benchmark", marker = "sys_platform == 'linux'" },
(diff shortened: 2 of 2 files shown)


FileSystemGuy:
@idevasena Just Finishing off the DLIO work you helped fix up!

idevasena (approved):
Looks good! Thank you @FileSystemGuy

PR #485, opened 2026-06-22
cli: require --data-dir for training datagen and run

FileSystemGuy:
## Summary
- Make `--data-dir` required for `mlpstorage training datagen` and `mlpstorage training run`. `datasize` and `configview` stay optional (datasize has no on-disk dataset; configview just renders the merged config).
- Update unit tests in `test_cli.py`, `test_cli_parser.py`, `test_parser_modes.py`, and `test_rules_parser_gating_consistency.py` to pass `--data-dir` for training `datagen`/`run` invocations.

## Motivation
Training `datagen` and `run` cannot operate without a dataset location. With `--data-dir` optional, omitting it failed deep in DLIO with an opaque error rather than at the CLI boundary.

## Notes on checkpointing
Checkpointing has no `--data-dir` (it uses `--checkpoint-folder`, already required for `run`) and no `datagen` subcommand, so no analogous change applied there.

## Test plan
- [x] `pytest tests/unit` — 1373 pass, 4 skipped (2 unrelated pre-existing version-string failures, 3 pre-existing import errors from missing optional deps).
- [ ] Manual smoke: `mlpstorage closed training unet3d datagen -np 4 file -rd /tmp` should now exit with `error: the following arguments are required: --data-dir/-dd`.


FileSystemGuy:
@russfellows Please approve and merge

russfellows (changes requested):
# Review

The current PR branch has the single-line parser-level `required=command in ("datagen", "run")` change, and validation still only checks `training run`. So the fix should move enforcement out of argparse and into post-YAML validation, then add object/config-file tests and clearer help text.

There are **3 comments**, but only **1 new regression / functional bug**.

There is no P1 bug because I did not find object-storage data corruption or a guaranteed broken object-storage command when `--data-dir` is supplied explicitly. The real must-fix is **P2**. The two **P3s** are lower-severity but still worth asking for because this PR changes CLI contract.

**Issue Count**
1. **P2 functional regression**: parser-level `required=True` prevents `--config-file` from supplying `data_dir`.
2. **P3 test gap**: no object-storage coverage for the new required `--data-dir` behavior.
3. **P3 UX/docs gap**: help text says “Filesystem location” even though object mode treats it as an object key prefix or URI.

Here are the specific code-change recommendations I’d make on the PR.

**1. Move Required Validation After YAML Overrides**
Do **not** make `--data-dir` required in argparse. Keep it optional at the parser layer so `--config-file` can populate it, then enforce it in post-parse validation.

In mlpstorage_py/cli/training_args.py, change:

```python
parser.add_argument(
    '--data-dir', '-dd',
    type=str,
    required=command in ("datagen", "run"),
    help="Filesystem location for data"
)
```

to:

```python
parser.add_argument(
    '--data-dir', '-dd',
    type=str,
    help=(
        "Dataset location. For file storage, this is a filesystem path. "
        "For object storage, this is an object key prefix or full object URI."
    )
)
```

Then in mlpstorage_py/validation_helpers.py, change:

```python
if command == 'run' and not getattr(args, 'data_dir', None):
    errors.append(ConfigurationError(
        "Missing required parameter: data-dir for training run",
        parameter="data_dir",
        suggestion="Generate data first with 'mlpstorage training datagen'",
        code=ErrorCode.CONFIG_MISSING_REQUIRED
    ))
```

to:

```python
if command in ('datagen', 'run') and not getattr(args, 'data_dir', None):
    errors.append(ConfigurationError(
        f"Missing required parameter: data-dir for training {command}",
        parameter="data_dir",
        suggestion=(
            "Specify --data-dir <path-or-prefix>. For object storage, use "
            "an object key prefix such as 'retinanet' or a full URI such as "
            "'s3://bucket/retinanet'."
        ),
        code=ErrorCode.CONFIG_MISSING_REQUIRED
    ))
```

This keeps the intended behavior — `datagen` and `run` require a dataset location — without breaking YAML config overrides.

**2. Add Regression Test For Config-File `data_dir`**
Add a test proving `data_dir` can still come from `--config-file`, especially for object storage.

A good test belongs near the existing config-file parser tests in tests/unit/test_cli.py:

```python
def test_training_datagen_data_dir_can_come_from_config_file(self, tmp_path):
    config_file = tmp_path / "object-config.yaml"
    config_file.write_text("data_dir: retinanet\n")

    test_args = [
        "mlpstorage",
        "closed",
        "training",
        "unet3d",
        "datagen",
        "--num-processes",
        "4",
        "--results-dir",
        "/tmp/results",
        "--config-file",
        str(config_file),
        "object",
    ]

    with patch("sys.argv", test_args):
        args = parse_arguments()

    assert args.command == "datagen"
    assert args.data_access_protocol == "object"
    assert args.data_dir == "retinanet"
```

This test would fail with the current PR because argparse exits before `apply_yaml_config_overrides()` runs.

**3. Add Object-Storage Explicit `--data-dir` Test**
Add a direct object-mode parser case so future changes do not accidentally assume filesystem-only paths.

```python
def test_training_datagen_object_accepts_object_key_prefix(self):
    test_args = [
        "mlpstorage",
        "closed",
        "training",
        "unet3d",
        "datagen",
        "--num-processes",
        "4",
        "--data-dir",
        "retinanet",
        "--results-dir",
        "/tmp/results",
        "object",
    ]

    with patch("sys.argv", test_args):
        args = parse_arguments()

    assert args.command == "datagen"
    assert args.data_access_protocol == "object"
    assert args.data_dir == "retinanet"
```

Optionally add the full URI variant too:

```python
def test_training_datagen_object_accepts_full_s3_uri(self):
    test_args = [
        "mlpstorage",
        "closed",
        "training",
        "unet3d",
        "datagen",
        "--num-processes",
        "4",
        "--data-dir",
        "s3://bucket/unet3d",
        "--results-dir",
        "/tmp/results",
        "object",
    ]

    with patch("sys.argv", test_args):
        args = parse_arguments()

    assert args.data_access_protocol == "object"
    assert args.data_dir == "s3://bucket/unet3d"
```

That said, I’d recommend docs continue to prefer the bare prefix form (`retinanet`, `unet3d/data`, etc.) because mlpstorage_py/benchmarks/dlio.py treats `data_dir` as `dataset.data_folder`, and object storage root/bucket is configured separately.

**Suggested PR Review Summary**
You can paste something like this into the PR:

```text
I agree with requiring a data location for training datagen/run, but I think the enforcement should move out of argparse and into post-parse validation. Today --config-file is applied after parser.parse_args(), so making --data-dir parser-required prevents configs from supplying data_dir. This is especially relevant for object-storage workflows where data_dir is often part of the object-storage config.

Suggested fix:
1. Remove required=command in ("datagen", "run") from the argparse definition.
2. Update validate_required_parameters() to require data_dir for training datagen and run after YAML overrides are applied.
3. Update help text to describe both file and object storage semantics.
4. Add tests for object mode with explicit --data-dir and for data_dir supplied by --config-file.
```

So: **3 suggestions, but only the first one is a must-fix functional regression**.

russfellows:
Curtis,

There was one new bug introduced, and a few documentation issues, I suggested changes in the PR.

—Russ

dslik (dismissed):
This changeset good to me.

**Edit** Looks like our reviews overlapped. @russfellows  I agree that we should not assume file system style paths, so your suggestion to parse this at the next deeper layer makes sense.

FileSystemGuy:
@russfellows @dslik I agree that this did not adequately address object.  I'll rework it.

FileSystemGuy:
Reworked per the review. Per-point summary:

**P2 (regression) — `--config-file` blocked by argparse-required.** Fixed by splitting enforcement by access protocol:
- File mode: checked in `cli_parser.parse_arguments` between `parse_args()` and YAML overrides, via `parser.error(...)`. Users still get the immediate argparse-style "the following arguments are required" UX.
- Object mode: checked in `validate_training_arguments` after YAML overrides, so `data_dir` supplied via `--config-file` works. Fails with a structured error message if neither source provides it.

This was @FileSystemGuy's call — keeping the fast-fail UX for the common file-storage path without doing a deeper rework to move YAML loading into the parser. Object-mode users lose the parse-time error but gain config-file flexibility.

**P3 (test gap) — object-storage coverage.** Added `TestTrainingDataDirEnforcement` in `tests/unit/test_cli_parser.py` covering:
- File mode + `datagen`/`run` without `--data-dir` → exits at parse time.
- Object mode + explicit `--data-dir` for bare prefix (`retinanet`), full URI (`s3://bucket/unet3d`), and nested prefix (`unet3d/data`) → all parse.
- Object mode + `data_dir` supplied only via `--config-file` → parses, with `args.data_dir` populated from YAML.
- Object mode + no `--data-dir` and no config-file source → exits with `EXIT_CODE.INVALID_ARGUMENTS`.

**P3 (help text).** Rewrote `--data-dir` help to: *"Dataset location. For file storage, this is a filesystem path. For object storage, this is an object key prefix or full object URI."*

One thing I didn't touch and want to flag: `_validate_required_params` in `mlpstorage_py/validation_helpers.py` reads `getattr(args, 'program', None)`, but the actual namespace attribute is `args.benchmark` — so the existing `data_dir` check at line 127 (and the other `program ==` checks in that function) currently never fires. Looks like leftover from a rename. Happy to file a separate issue/PR for that since it's outside this PR's scope.

FileSystemGuy:
@russ @dslik Just a ping to let you know that this was "fixed"....

russfellows (approved):
Approving, since Curtis fixed the issue.

Issue #475, opened 2026-06-18
[DLIO] Persistent Workers Cause Stale File Shards and Cached Reads


wolfgang-desalvador:
## Summary

When `persistent_workers=True` (the default for `num_workers > 0`), PyTorch DataLoader worker processes are kept alive across epochs. This causes two compounding issues:

1. **Stale file assignments**: Workers capture a pickled snapshot of `ConfigArguments` at spawn time and never receive updates from `reconfigure()`, so cross-epoch file shuffling and resharding have no effect.
2. **Skipped I/O**: The local filesystem reader's in-memory cache (`self._local_cache`) persists across epochs, so files read in epoch 1 are never re-read from storage — subsequent epochs measure memory bandwidth, not storage throughput.

Both issues stem from the same root cause: persistent workers retain process-level state that should be reset each epoch.

### Stale file assignments

In [`dlio_benchmark/data_loader/torch_data_loader.py`](https://github.com/mlcommons/DLIO_local_changes/blob/main/dlio_benchmark/data_loader/torch_data_loader.py), the `TorchDataset` class serializes the config at construction:

```python
class TorchDataset(Dataset):
    def __init__(self, ...):
        args = ConfigArguments.get_instance()
        self.serial_args = pickle.dumps(args)    # snapshot at construction

    def worker_init(self, worker_id):
        pickle.loads(self.serial_args)           # restores stale snapshot
```

With `persistent_workers=True`, workers are spawned once and `worker_init` runs only at first spawn. The `serial_args` snapshot is never updated, so workers always see epoch 1's `file_list_train`, `train_file_map`, `train_global_index_map`, etc. — even after `reconfigure()` shuffles or reshards files in the main process.

`persistent_workers=True` is set unconditionally in three places in `torch_data_loader.py` (lines 558, 592, and 627):

```python
if torch.__version__ != '1.3.1':
    kwargs['persistent_workers'] = True
```

### Skipped I/O (consequence of persistence)

In [`dlio_benchmark/reader/_local_fs_iterable_mixin.py`](https://github.com/mlcommons/DLIO_local_changes/blob/main/dlio_benchmark/reader/_local_fs_iterable_mixin.py), the `_localfs_ensure_cached` method checks whether a file is already in the worker's in-memory cache:

```python
def _localfs_ensure_cached(self, filename: str) -> None:
    if filename not in self._local_cache:       # <-- skips read if cached
        ...
```

[...]

Issue #28, opened 2023-05-27
Update DLRM mentions in readme, etc.


TheKanter:
First of all, the https://github.com/mlcommons/storage/blob/main/README.md is beautiful and really nicely done!

Since DLRM isn't in first version, so rather than saying "todo", can we explicitly indicate that?

johnugeorge:
Created: https://github.com/mlcommons/storage/pull/30

johnugeorge:
This is resolved with #30

PR #492, opened 2026-06-23
v3.0.17: --drop-caches-timeout-seconds CLI + KV cache wrapper consolidation (#498)

### LINKED ISSUE #487: Could not flush page cache between epochs warning
Flush page cache command times out during a training run.  Our system requires more than 30 seconds to flush caches.

Output from an mlpstorage closed training retinanet run file command:

[OUTPUT] 2026-06-20T13:34:43.831067 Worker pre-warm complete (8 workers spawned)
[WARNING] Could not flush page cache between epochs: Command '['sudo', '-n', 'sh', '-c', 'echo 3 > /proc/sys/vm/drop_caches']' timed out after 30 seconds. Per-epoch reads may be served from the OS buffer cache, inflating throughput numbers. To enable, configure passwordless sudo for `sh -c 'echo 3 > /proc/sys/vm/drop_caches'`.
[OUTPUT] 2026-06-20T13:35:13.865508 Starting epoch 1: 32641 steps expected

### LINKED ISSUE #498: Clarification on --config vs OPTION_PARAMS in mlperf_wrapper.py for multi-client runs
I have a question regarding the behavior of mlperf_wrapper.py when running multi-client (MPI) KV Cache benchmarks.

Context:
In the current mlperf_wrapper.py script, when I specify --option 1/2/3, the wrapper constructs the CLI command by explicitly passing parameters from OPTION_PARAMS (e.g., --model, --num-users, --duration), while also accepting a user-provided --config YAML file.

According to the code:

python
cmd = [
    ...
    '--config', config_path,
    '--model', params['model'],
    '--num-users', str(params['num-users']),
    '--duration', str(params['duration']),
    ...
]
Since CLI arguments take higher precedence than config file values in kv-cache.py, this means the parameters defined in OPTION_PARAMS always override any identical fields (e.g., model, num-users) written in the user-specified --config YAML.

My questions are:

1、In multi-client co-running scenarios (e.g., multiple ranks/nodes), is the --config parameter intended only for specifying model‑related parameters?

2、Is OPTION_PARAMS designed to be a mandatory enforcement for MLPerf v3.0 submission compliance? If so, is the official recommendation that users should only rely on the --option flag to select the predefined workload, and use --config solely for storage-tuning knobs that are not covered by OPTION_PARAMS?

I want to make sure I understand the official design philosophy to avoid misusing the wrapper in our multi-client test environment. Any clarification would be greatly appreciated.

Thank you for your great work on this benchmark suite!

### LINKED ISSUE #500: KVCache [--num-processes NUM_PROCESSES] doesn't affect the number of processe per host
The number of hosts is set by -- hosts and this should control the number of processes per host, but doesn't appear to have any effect.

I need to figure out why and then fix it.

FileSystemGuy:
## Summary

Closes #487, #498.

> Note: PR #501 (now merged in here) was originally framed as closing both #498 and #500. dslik re-tested and confirmed it did not fix #500 — that issue is about `--num-processes` being a dead flag (one instance per host regardless), which is a separate root cause from the WORKLOAD_PARAMS squash. #500 stays open and will land in a follow-up PR.

This PR bundles three related changes that ship together as `mlpstorage 3.0.17`:

1. **`training run --drop-caches-timeout-seconds`** — operator knob plumbed to DLIO via `DLIO_DROP_CACHES_TIMEOUT` (mlpstorage-side of #487 / DLIO PR #28). Original PR intent.
2. **KV cache wrapper consolidation (#498)** — `WORKLOAD_PARAMS` moved to `mlpstorage_py.benchmarks.kvcache`; `mlperf_wrapper.py` becomes a pure rank-aware launcher. Merged in via PR #501.
3. **Version + DLIO pin bump** — `3.0.16 → 3.0.17`. DLIO source switches from a fixed `rev` to `branch = "main"`; reproducibility is still anchored by the hash recorded in `uv.lock`. Re-resolve to the latest DLIO commit with `uv lock --upgrade-package dlio-benchmark` — no `pyproject.toml` edit needed.

---

## Part 1 — `--drop-caches-timeout-seconds` (closes #487, mlpstorage-side)

mlpstorage-side half of #487. The DLIO half is mlcommons/DLIO_local_changes#28, which adds the `DLIO_DROP_CACHES_TIMEOUT` env var so large-RAM hosts can raise the per-call timeout for DLIO's per-epoch page-cache flush.

This adds a deployment knob to the **training `run` subcommand only** and plumbs the value through to DLIO via that env var.

### Scope

- **Subcommand**: training `run` only. `datasize` / `datagen` / `configview` do not invoke the per-epoch flush loop.
- **Benchmark**: training only. Checkpointing, VectorDB, and KV Cache do not use the auto-flush code path.
- **Mode**: all (closed/open/whatif). It's a deployment knob, not a submission tunable, matching the existing pattern for `--dlio-bin-path`.

### Plumbing

1. **CLI** (`mlpstorage_py/cli/training_args.py`): argparse validates positive int (>= 1) via a `_positive_int` type. Default `None` means no override; DLIO falls back to its built-in 30s.
2. **Env-var set** (`mlpstorage_py/benchmarks/dlio.py:TrainingBenchmark.__init__`): when the flag is provided, set `os.environ['DLIO_DROP_CACHES_TIMEOUT']` to the stringified value. `subprocess.Popen` inherits parent env, so single-process invocations see it.
3. **MPI forwarding** (`generate_dlio_command()`): when the env var is set and `exec_type == MPI`, append `-x DLIO_DROP_CACHES_TIMEOUT` to the MPI prefix.

### Operator usage

```bash
mlpstorage closed training retinanet run file \
    --client-host-memory-in-gb 64 --num-accelerators 4 \
    --accelerator-type b200 --data-dir ... --results-dir ... \
    --drop-caches-timeout-seconds 300
```

---

## Part 2 — KV cache wrapper consolidation (closes #498)

`kv_cache_benchmark/mlperf_wrapper.py` was unconditionally injecting `WORKLOAD_PARAMS` as explicit CLI flags on every `kv-cache.py` invocation, overriding the user's `--config` YAML and CLI flags **even in OPEN mode**. CLOSED/OPEN gating already lives in `mlpstorage_py/cli/kvcache_args.py`, so the wrapper-side squashing was redundant in CLOSED and actively harmful in OPEN.

- `WORKLOAD_PARAMS` now lives only in `mlpstorage_py.benchmarks.kvcache` — single source of truth for what option N means at the kv-cache.py level.
- New `KVCacheBenchmark._build_option_kvcache_args(option, is_closed)` emits per-option flags: verbatim in CLOSED, with user CLI flags superseding per-option defaults in OPEN. `max-concurrent-allocs` always stays at the option's mandated value (no OPEN CLI flag).
- `mlperf_wrapper.py` reduced to a pure rank-aware launcher: `--rank-output-base`, `--rank-cache-base`, `--seed-base`, `--start-delay`, `--end-delay`; everything else forwards verbatim to `kv-cache.py`. Per-rank `--seed`/`--output`/`--cache-dir` are appended last so argparse store-action takes them over forwarded duplicates. `allow_abbrev=False` prevents a forwarded `--seed` from collapsing into `--seed-base`.

---

## Part 3 — Version + DLIO pin

- `version`: `3.0.16` → `3.0.17`
- `pyproject.toml` switches DLIO source from `rev = "<hash>"` to `branch = "main"`. The team owns `DLIO_local_changes` and merges into `main` only intentional changes, so tracking `main` is safe.
- `uv.lock` records the resolved hash for reproducibility. Current pin: `3667ed1` (latest `main`, includes DLIO PR #28).
- To pull in future DLIO commits: `uv lock --upgrade-package dlio-benchmark`.

---

## Test plan

- [x] `pytest tests/unit/test_drop_caches_timeout_cli.py` — 25 cases pass (flag scope, value validation, env-var plumbing, MPI forwarding, end-to-end).
- [x] `pytest tests/unit/test_mlperf_wrapper.py tests/unit/test_benchmarks_kvcache.py` — 85 cases pass covering `WORKLOAD_PARAMS` move, `_build_option_kvcache_args` CLOSED/OPEN behavior, forwarded-arg passthrough, per-rank override ordering, end-to-end wrapper-cmd assertions (incl. OPEN option-3 70B fallback and user `--config` path supersession).
- [x] `pytest tests/unit tests/integration/test_benchmark_flow.py::TestKVCacheRunIntegration --ignore=tests/unit/test_benchmarks_base.py --ignore=tests/unit/test_parquet_reader.py --ignore=tests/unit/test_vdb_modular_fake_backend.py` — 1,442 pass, 4 skipped (pre-existing missing-dep skips), 0 failures.
- [ ] Manual: KV cache OPEN run with `--config user.yaml` to confirm user values reach `kv-cache.py`.
- [ ] Manual: training run with `--drop-caches-timeout-seconds 300` to confirm `DLIO_DROP_CACHES_TIMEOUT=300` reaches DLIO ranks.

### DIFF
--- a/kv_cache_benchmark/mlperf_wrapper.py
+++ b/kv_cache_benchmark/mlperf_wrapper.py
@@ -1,10 +1,12 @@
 #!/usr/bin/env python3
-"""MPI-rank-aware wrapper for MLPerf KV Cache benchmark.
+"""MPI-rank-aware launcher for the MLPerf KV Cache benchmark.
 
 Invoked by mpirun per-rank; reads OMPI_COMM_WORLD_RANK (OpenMPI) or PMI_RANK
-(MPICH) to determine this rank's index. Computes per-rank seed, output
-directory, and cache directory, then invokes kv-cache.py with the fixed MLPerf
-v3.0 parameters for the specified option.
+(MPICH) to determine this rank's index. The wrapper does NOT encode any
+workload parameters — it only computes per-rank values (seed, output file,
+cache directory) and forwards everything else to kv-cache.py. The caller
+(mlpstorage_py.benchmarks.kvcache.KVCacheBenchmark) owns the per-option
+workload parameter set and CLOSED/OPEN enforcement.
 """
 
 import argparse
@@ -17,40 +19,6 @@ from pathlib import Path
 
 BASE_SEED = 42
 TEST_DELAY = 90
-
-# MLPerf v3.0 fixed parameters per option.
-# All numeric values stored as int/float; converted to str when building cmd.
-# 'generation-mode' is ALWAYS 'none' for MLPerf compliance — do NOT rely on
-# kv-cache.py defaults; the default may change in future versions.
-WORKLOAD_PARAMS = {
-    1: {
-        'model': 'llama3.1-8b',
-        'num-users': 200,
-        'duration': 300,
-        'gpu-mem-gb': 0,
-        'cpu-mem-gb': 0,
-        'max-concurrent-allocs': 16,
-        'generation-mode': 'none',
-    },
-    2: {
-        'model': 'llama3.1-8b',
-        'num-users': 100,
-        'duration': 300,
-        'gpu-mem-gb': 0,
-        'cpu-mem-gb': 4,
-        'max-concurrent-allocs': 16,
-        'generation-mode': 'none',
-    },
-    3: {
-        'model': 'llama3.1-70b-instruct',
-        'num-users': 70,
-        'duration': 300,
-        'gpu-mem-gb': 0,
-        'cpu-mem-gb': 0,
-        'max-concurrent-allocs': 4,
-        'generation-mode': 'none',
-    },
-}
 
 
 def get_rank() -> int:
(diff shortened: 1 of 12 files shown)


FileSystemGuy:
Resolves #475

FileSystemGuy:
@dslik I'm sorry that the kvcache args-suppression cleanup got bundled into this PR, the branches got merged in the opposite order that they were intended, but all that work is there and a slew of new unit tests claim that it all works.

@idevasena @russfellows @dslik Any one of you could review and approve this, please.

russfellows (approved):
New unit tests should verify proper operation, so LGTM.

PR #493, opened 2026-06-23
Add rules for picking subsets in runs and checkpointing

FileSystemGuy:
Added rules for runPickingSubsets and checkpointingPickingSubsets.

In v2.0 picking a consecutive subset of training runs was permissible, but that rule didn't make it into v3.0's Rules.md file.


idevasena (dismissed):

russfellows (approved):
Good to go.

PR #497, opened 2026-06-23
Fix #463: apply nested index_params and default-valued config keys

### LINKED ISSUE #463: vdb_benchmark : indexing parameter values from config files are not getting used
Issue 1: in default.yaml and 10m.yaml, for DiskANN as index type, indexing parameters `max_degree` and `search_list_size` are expected instead of M and ef_construction.  

Issue 2: yaml config file parser having issue in parsing second level nested parameters under index_params:
Hence even after changing default config file as below, datagen command still uses default values from load_db.py.
```
+++ b/configs/vectordbbench/default.yaml
@@ -16,11 +16,11 @@ dataset:
   vector_dtype: FLOAT_VECTOR

 index:
-  index_type: DISKANN
+  index_type: HNSW
   metric_type: COSINE
   index_params:
-    M: 64
-    ef_construction: 200
+    M: 32
+    ef_construction: 100
```
Datagen command output after above changes:
```
blr-host01:~/psangle/storage (main)$ ./mlpstorage vectordb datagen --host 127.0.0.1 --port 19530 --config default --force --results-dir /tmp/vdb_results
warning: The `UV_NATIVE_TLS` environment variable is deprecated and will be removed in a future release. Use `UV_SYSTEM_CERTS` instead.
⠋ Validating environment... 0:00:002026-06-17 12:14:55|INFO: Environment validation passed
2026-06-17 12:14:55|STATUS: Benchmark results directory: /tmp/vdb_results/vector_database/datagen/20260617_121455
2026-06-17 12:14:56|INFO: Created benchmark run: vector_database_datagen_20260617_121455
2026-06-17 12:14:56|STATUS: Verifying benchmark run for vector_database_datagen_20260617_121455
2026-06-17 12:14:56|STATUS: Open: [OPEN] VectorDB benchmark is in preview status - not accepted for closed submissions (Parameter: benchmark_status)
2026-06-17 12:14:56|STATUS: Benchmark run qualifies for OPEN category ([RunID(program='vector_database', command='datagen', model=None, run_datetime='20260617_121455')])
2026-06-17 12:14:56|WARNING: Running the benchmark without verification for open or closed configurations. These results are not valid for submission. Use --open or --closed to specify a configuration.
2026-06-17 12:14:56|STATUS: Instantiated the VectorDB Benchmark...
warning: The `UV_NATIVE_TLS` environment variable is deprecated and will be removed in a future release. Use `UV_SYSTEM_CERTS` instead.
2026-06-17 12:14:56,365 - INFO - Connected to Milvus server at 127.0.0.1:19530
2026-06-17 12:14:56,365 - WARNING - FLOAT16 data type not available in this version of pymilvus. Using FLOAT_VECTOR instead.
2026-06-17 12:14:56,375 - INFO - Dropped existing collection: mlps_1m_1shards_1536dim_uniform
2026-06-17 12:14:56,385 - INFO - Created collection 'mlps_1m_1shards_1536dim_uniform' with 1536 dimensions and 1 shards
2026-06-17 12:14:56,385 - INFO - Creating index with parameters: **{'index_type': 'HNSW', 'metric_type': 'COSINE', 'params': {'M': 16, 'efConstruction': 200}}**
```

Refer last line of above log snippet, index type is reflecting updated value as `HNSW` but indexing parameters `M` and `ef_construction` are NOT changed as per config.yaml changes. With other combinations of index type and parameter values as well issue persists.
Please cross check in your setup and align indexing parameter use as per config files under `vdb_benchmark/vdbbench/configs/`.

idevasena:
Fix #463: vdb_benchmark config index params (nested + default-valued) not applied:

1. Root cause 1 — nested index_params: is silently dropped. `merge_config_with_args` loops for section, params in config.items() then for key, value in params.items(), i.e. exactly two levels. When tuning params are nested as index: { index_params: { M, ef_construction } }, the inner loop sees key="index_params" with a dict value. index_params is not an argparse destination, so the whole block is skipped and M/ef_construction stay at their argparse defaults (16/200). That is precisely the log line in the issue: index_type updates to HNSW (it sits one level up, so it merges), but M/efConstruction revert to 16/200.

2. Root cause 2 — config values equal to the argparse default are ignored. The merge only writes a config value when the arg is None or is_default[key] is True. It can't distinguish "user typed --M 16" from "config asked for 16," and a config value that happens to equal the default is treated as still-default and dropped. This bites flat configs too.

Note main's flat configs (M/ef_construction directly under index:) merge correctly today because they aren't nested — so the bug surfaces whenever someone uses the index_params: nesting shown in the issue, or sets a value matching a default.

### Fix
In `vdb_benchmark/vdbbench/config_loader.py`:

1. Added `_flatten_config()` that recursively lifts every leaf key to the top level, so index_params.M becomes M.
Reworked precedence to: explicit CLI flag > YAML config > argparse default. A YAML value is now always applied unless the user explicitly set that flag on the CLI (detected via the existing args.is_default map). Unknown config keys (e.g. max_receive_message_length) are ignored exactly as before, and callers that consume those directly from the config dict are unaffected.

This single change fixes every caller (`load_vdb`, `simple_bench`, `enhanced_bench`, `mpi_wrapper`, `compact_and_watch`), since they all share this function. Added 8 regression tests in 1tests/tests/test_issue_463_nested_index_params.py`.

### DIFF
--- a/configs/vectordbbench/10m.yaml
+++ b/configs/vectordbbench/10m.yaml
@@ -19,8 +19,9 @@ index:
   index_type: DISKANN
   metric_type: COSINE
   index_params:
-    M: 64
-    ef_construction: 200
+    max_degree: 64
+    search_list_size: 200
 
 workflow:
   compact: True
+
--- a/configs/vectordbbench/default.yaml
+++ b/configs/vectordbbench/default.yaml
@@ -19,8 +19,8 @@ index:
   index_type: DISKANN
   metric_type: COSINE
   index_params:
-    M: 64
-    ef_construction: 200
+    max_degree: 64
+    search_list_size: 200
 
 workflow:
   compact: True
@@ -32,3 +32,4 @@ benchmark:
   report_count: 100
   recall_k: 10
   search_limit: 10
+
--- a/vdb_benchmark/tests/tests/test_issue_463_nested_index_params.py
+++ b/vdb_benchmark/tests/tests/test_issue_463_nested_index_params.py
@@ -127,3 +127,63 @@ def test_empty_config_is_noop():
     out = merge_config_with_args({}, args)
     assert out.M == 16 and out.index_type == "DISKANN"
 
+
+def test_cross_section_same_leaf_does_not_collide():
+    """
+    dataset.batch_size and benchmark.batch_size both exist in the shipped
+    configs. They mean different things and must not collapse into one value.
+    The earlier section (dataset) wins for the shared argparse dest.
+
+    Regression for PR #497 review (ram-sangle): _flatten_config must not let a
+    later section silently overwrite an earlier same-named leaf.
+    """
+    cfg = yaml.safe_load("""
+    dataset:
+      batch_size: 10
+    benchmark:
+      batch_size: 1
+    """)
+    flat = _flatten_config(cfg)
+    assert flat["batch_size"] == 10  # dataset (earlier) wins, not benchmark
+
+
+def test_intra_section_index_params_collision_raises():
+    """
+    A collision *within* one section (a section-level key duplicated by a key
+    nested under index_params) is a genuine mistake and must surface, not be
+    silently resolved.
+    """
+    import pytest
+
+    cfg = yaml.safe_load("""
+    index:
+      index_type: HNSW
+      M: 16
+      index_params:
+        M: 64
+    """)
+    with pytest.raises(ValueError):
+        _flatten_config(cfg)
+
+
+def test_diskann_config_uses_diskann_params():
+    """
+    The shipped DISKANN configs must declare max_degree/search_list_size, not
+    HNSW's M/ef_construction. Regression for PR #497 review (ram-sangle).
+    """
+    cfg = yaml.safe_load("""
+    index:
+      index_type: DISKANN
+      metric_type: COSINE
+      index_params:
+        max_degree: 64
+        search_list_size: 200
+    """)
+    args = merge_config_with_args(cfg, _make_args())
+    assert args.index_type == "DISKANN"
+    assert args.max_degree == 64
+    assert args.search_list_size == 200
+    # HNSW params remain at their defaults, untouched.
+    assert args.M == 16
+    assert args.ef_construction == 200
+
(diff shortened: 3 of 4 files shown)


ram-sangle:
@idevasena the fix looks good to me, consider updating diskann indexing parameters in `configs/vectordbbench/default.yaml` and `configs/vectordbbench/10m.yaml` as well instead of using HNSW parameters.

```
  index_params:
    max_degree: 64
    search_list_size: 200
```

ram-sangle (commented):
Please check any possibility of key collision as _flatten_config() walks all sections. If in future `database.batch_size` and `dataset.batch_size` both exist, the last value will be retained. Recursing within a specific section might work here.

idevasena:
> @idevasena the fix looks good to me, consider updating diskann indexing parameters in `configs/vectordbbench/default.yaml` and `configs/vectordbbench/10m.yaml` as well instead of using HNSW parameters.
> 
> ```
>   index_params:
>     max_degree: 64
>     search_list_size: 200
> ```

@ram-sangle Good catch. Done. Both `configs/vectordbbench/default.yaml``and `configs/vectordbbench/10m.yaml` declared index_type: DISKANN but carried HNSW's `M/ef_construction` under `index_params:` — and since load_vdb reads `max_degree/search_list_size` for DISKANN, those keys were dead and the index was silently falling back to the argparse defaults. Replaced them with:
```
yamlindex_params:
  max_degree: 64
  search_list_size: 200
```
in both files. Added `test_diskann_config_uses_diskann_params` to confirm these propagate through `merge_config_with_args` and that the HNSW params stay untouched.

idevasena:
> Please check any possibility of key collision as _flatten_config() walks all sections. If in future `database.batch_size` and `dataset.batch_size` both exist, the last value will be retained. Recursing within a specific section might work here.

Agreed, and this turned out to be live rather than hypothetical — `configs/vectordbbench/default.yaml` already has both `dataset.batch_size: 10` and `benchmark.batch_size: 1`, and the global flatten was collapsing them last-wins, so the datagen insert batch size was silently becoming 1.

Reworked per your suggestion to flatten per section rather than across the whole document:

Cross-section duplicate leaves now keep the first occurrence in document order (so dataset wins over benchmark for the shared `batch_size` dest) and emit a WARNING so the duplication is visible.
An intra-section collision — e.g. a key set at the section level and again under i`ndex_params:` — now raises ValueError instead of being silently resolved.

Added `test_cross_section_same_leaf_does_not_collide` and `test_intra_section_index_params_collision_raises` to lock this in.

russfellows (approved):
Approved.

PR #501, opened 2026-06-24
fix(kvcache): consolidate per-option workload params in mlpstorage (#498, #500)

### LINKED ISSUE #498: Clarification on --config vs OPTION_PARAMS in mlperf_wrapper.py for multi-client runs

FileSystemGuy:
## Summary

Closes #498 and #500. Both issues report the same root cause: `kv_cache_benchmark/mlperf_wrapper.py` was unconditionally injecting `WORKLOAD_PARAMS` as explicit CLI flags on every `kv-cache.py` invocation, overriding the user's `--config` YAML and CLI flags **even in OPEN mode**. CLOSED/OPEN gating already lives in `mlpstorage_py/cli/kvcache_args.py`, so the wrapper-side squashing was redundant in CLOSED and actively harmful in OPEN.

- `WORKLOAD_PARAMS` now lives only in `mlpstorage_py.benchmarks.kvcache` — single source of truth for what option N means at the kv-cache.py level.
- New `KVCacheBenchmark._build_option_kvcache_args(option, is_closed)` emits per-option flags: verbatim in CLOSED, with user CLI flags superseding per-option defaults in OPEN. `max-concurrent-allocs` always stays at the option's mandated value because it has no OPEN CLI flag.
- `mlperf_wrapper.py` reduced to a pure rank-aware launcher: takes `--rank-output-base`, `--rank-cache-base`, `--seed-base`, `--start-delay`, `--end-delay`; everything else forwards verbatim to `kv-cache.py`. Per-rank `--seed`/`--output`/`--cache-dir` are appended last so argparse store-action takes them over forwarded duplicates. `allow_abbrev=False` prevents a forwarded `--seed` from collapsing into `--seed-base`.

## Base branch note

This PR is currently based on `feat/drop-caches-timeout-cli` (PR #492) to keep the diff clean. Once #492 lands, please retarget the base to `main`.

## Test plan

- [x] `tests/unit/test_mlperf_wrapper.py` — 14 tests covering rank detection, new wrapper API, forwarded-arg passthrough, per-rank override ordering, and `WORKLOAD_PARAMS` removal
- [x] `tests/unit/test_benchmarks_kvcache.py` — 69 tests including new `TestWorkloadParamsConstant`, `TestBuildOptionKvcacheArgs` (CLOSED verbatim, OPEN user-override, max-concurrent-allocs invariant), and `TestWrapperCommandForwardsPerOptionArgs` (end-to-end wrapper-cmd assertions for CLOSED and OPEN)
- [ ] Manual verification on a real KV cache run with `--config` + open mode to confirm user YAML values flow through

### DIFF
--- a/kv_cache_benchmark/mlperf_wrapper.py
+++ b/kv_cache_benchmark/mlperf_wrapper.py
@@ -1,10 +1,12 @@
 #!/usr/bin/env python3
-"""MPI-rank-aware wrapper for MLPerf KV Cache benchmark.
+"""MPI-rank-aware launcher for the MLPerf KV Cache benchmark.
 
 Invoked by mpirun per-rank; reads OMPI_COMM_WORLD_RANK (OpenMPI) or PMI_RANK
-(MPICH) to determine this rank's index. Computes per-rank seed, output
-directory, and cache directory, then invokes kv-cache.py with the fixed MLPerf
-v3.0 parameters for the specified option.
+(MPICH) to determine this rank's index. The wrapper does NOT encode any
+workload parameters — it only computes per-rank values (seed, output file,
+cache directory) and forwards everything else to kv-cache.py. The caller
+(mlpstorage_py.benchmarks.kvcache.KVCacheBenchmark) owns the per-option
+workload parameter set and CLOSED/OPEN enforcement.
 """
 
 import argparse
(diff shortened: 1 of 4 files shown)


dslik (approved):
Reviewed the code. Will merge and run additional tests.

PR #511, opened 2026-06-24
fix(rules): stop reportgen crash on None system_info; share TOOL_INJECTED_PARAMS with submission_checker (#503)

### LINKED ISSUE #503: reportgen validation incorrectly marks all runs as INVALID due to 3 bugs
When running mlpstorage reports reportgen --results-dir <dir>, all runs are reported as INVALID due to the following issues:
                                                                                                                                                                                               1. check_num_files_train crashes with 'NoneType' object has no attribute 'total_memory_bytes'

  When loading runs from result files on disk (reportgen path), benchmark_run.system_info is None. In mlpstorage_py/rules/run_checkers/training.py:85-92, this None value is passed as
  cluster_information to calculate_training_data_size(). In mlpstorage_py/rules/utils.py:51-52, the function directly accesses cluster_information.total_memory_bytes without a null check,
  causing an AttributeError. The exception is caught and the run is marked INVALID.

  2. dataset.skip_listing flagged as disallowed parameter override

  TrainingBenchmark.__init__ calls _apply_skip_listing_params() (mlpstorage_py/benchmarks/dlio.py:243-275) which injects dataset.skip_listing = True into self.params_dict. This flows into
  override_parameters metadata. However, dataset.skip_listing is not listed in CLOSED_ALLOWED_PARAMS (line 19-31) or OPEN_ALLOWED_PARAMS in mlpstorage_py/rules/run_checkers/training.py, nor
   in the submission checker's allowed_params in mlpstorage_py/submission_checker/checks/training_checks.py. The validation then flags it as a disallowed override.

  3. dataset.listing_validation_interval flagged as disallowed parameter override

  Same mechanism as #2 — _apply_skip_listing_params() also injects dataset.listing_validation_interval into params_dict, but it's missing from all allowed parameter lists.

  Note: These two parameters are injected automatically by the code, not by the user's --params argument.

  The combination of these bugs causes all 8 runs (6 training runs, 1 datasize, 1 datagen) in our results directory to be marked INVALID with no usable metrics.

FileSystemGuy:
## Summary

Closes #503.

Issue #503 reported three bugs that all marked reportgen-loaded runs as INVALID for the wrong reasons. Two of those (`dataset.skip_listing` and `dataset.listing_validation_interval` flagged as disallowed user overrides) had already been fixed on the in-process verifier path in commit `0b3d370` (PR #496 / issue #494) two days before #503 was filed — but the submission_checker side never got the same fix, so the two checkers gave divergent verdicts for the same metadata. Bug 1 (`AttributeError` on `cluster_information.total_memory_bytes` when loading from disk) is genuinely new.

This PR fixes both real problems and adds tests guarding the shared invariant.

## Part 1 — Bug 1: `check_num_files_train` crash on missing `system_info`

Reportgen loads benchmark runs from on-disk metadata. When that metadata doesn't carry a serialized `cluster_information` block, `BenchmarkRun.system_info` is `None`. The check then called:

```python
required_num_files, _, _ = calculate_training_data_size(
    None, self.benchmark_run.system_info, ...
)
```

Inside `utils.py:85-86`:
```python
if not args:
    total_mem_bytes = cluster_information.total_memory_bytes  # AttributeError on None
```

The verifier's per-check exception wrapper caught the `AttributeError` and marked the entire run INVALID. Every reportgen-loaded training run hit this — the user's "all 8 runs INVALID" symptom.

Fix:
- `mlpstorage_py/rules/utils.py:84-100` — guard the None case and raise a descriptive `ValueError` instead of crashing on attribute access.
- `mlpstorage_py/rules/run_checkers/training.py:check_num_files_train` — catch that specific `ValueError`, log a warning naming the check, and return `None` (skip). The check requires live cluster info; on-disk metadata that doesn't preserve it cannot be validated against the 5×memory rule, so skipping with a clear log message is the right action, not marking INVALID.

## Part 2 — Bugs 2 & 3: submission_checker drift from run_checkers

`mlpstorage_py/rules/run_checkers/training.py:51-63` defines `TOOL_INJECTED_PARAMS` — the dotted-keys that mlpstorage writes into `params_dict` on the user's behalf (`dataset.skip_listing`, `dataset.listing_validation_interval`, `dataset.data_folder` when derived from `--data-dir`, the object-storage backend keys). The in-process verifier's `check_allowed_params` skips them so they don't count as user overrides.

`mlpstorage_py/submission_checker/checks/training_checks.py:545-579, 581-636` never got the same exemption. Rules 3.6.2 (closed) and 3.6.3 (open) flagged any `params_dict` key not in their flat `allowed_params` set, so they reported `dataset.skip_listing` as a CLOSED-disallowed parameter override.

Fix: import the same `TOOL_INJECTED_PARAMS` frozenset from the canonical source and exempt those keys in both 3.6.2 and 3.6.3. Single source of truth — any future tool-injected key added in `run_checkers/training.py` is automatically covered by the submission_checker too.

## Test plan

- [x] **`test_rules_calculations.py::test_raises_value_error_when_both_args_and_cluster_info_are_none`** — `calculate_training_data_size(args=None, cluster_information=None, …)` raises `ValueError` mentioning `cluster_information`, not `AttributeError`.
- [x] **`test_rules_checkers.py::test_check_num_files_train_skips_when_system_info_missing`** — `check_num_files_train` returns `None` and logs a warning naming the check when `system_info` is `None`. Confirms the AttributeError → INVALID-cascade is broken.
- [x] **`test_training_check_tool_injected_params.py`** — parametrized over every key in `TOOL_INJECTED_PARAMS`: rules 3.6.2 and 3.6.3 do NOT flag them. Regression guard confirms genuinely disallowed params (`not.a.real.param`) are still rejected.
- [x] `pytest tests/unit tests/integration mlpstorage_py/tests/test_training_check_tool_injected_params.py` (no `--ignore`, no selective collection) → **1,637 passed, 15 skipped, 12 deselected, 0 failures**.

## Notes for the reviewer

- Includes a cherry-picked test-infrastructure fix (`a2fecab`) from PR #509 — the pyarrow stub guard and the `config.py::_resolve_default_results_dir` extraction — so this branch's full test suite runs cleanly without `--ignore`. If PR #509 lands first, that commit becomes a clean no-op here; if this one lands first, PR #509's version is a no-op there.
- The TOOL_INJECTED_PARAMS share creates a new cross-module import (`mlpstorage_py.submission_checker.checks.training_checks` → `mlpstorage_py.rules.run_checkers.training`). No circularity risk — `rules` is a sibling subpackage and doesn't import from `submission_checker`.

### DIFF
--- a/Rules.md
+++ b/Rules.md
@@ -16,6 +16,7 @@
     * [3.6. Training OPEN versus CLOSED Options](#36-training-open-versus-closed-options)
 * [4. Validating the Checkpointing Workloads](#4-validating-the-checkpointing-workloads)
     * [4.1. Checkpointing Sizing Options](#41-checkpointing-sizing-options)
+    * 
     * [4.2. Checkpointing Generation Options](#42-checkpointing-generation-options)
     * [4.3. Checkpointing Run Options](#43-checkpointing-run-options)
     * [4.4. Checkpointing Access Via POSIX API Options](#44-checkpointing-access-via-posix-api-options)
@@ -118,6 +119,8 @@ configuration of storage system and to link together those results with the .pdf
 
 2.1.18. **runTimestampGap** --  The timestamp (the day and time) represented by the name of each *timestamp directory* must be separated by less than the duration of a single *timestamp directory* from it's neighboring *timestamp directories*.  Ie: the gap between a consecutive pair of *timestamp directories* must be short enough that we can be sure that there was no benchmark activity between them.
 
+2.1.18a.  **runPickingSubsets** -- It is permissible to execute a large number of runs (all consecutive) and then delete the *timestamp directories* for all but 6 consecutive runs in the middle.  This differs from "cherry picking" in that these runs must be consecutive, see **runTimestampGap**.
+
 2.1.19. **runFiles** --  Within each *timestamp directory* within the "run" *phase*, there must exist the following files: "training_run.stdout.log", "training_run.stderr.log" file, "*output.json, "*per_epoch_stats.json", "*summary.json", and "dlio.log", plus a subdirectory named "dlio_config".  These names are case-sensitive.
 
 2.1.20. **runDlioConfig** --  The "dlio_config" subdirectory in each *timestamp directory* must contain the following list of files, and nothing else: "config.yaml", "hydra.yaml", and "overrides.yaml".  These names are case-sensitive.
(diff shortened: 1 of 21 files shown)


russfellows (approved):
Good to go.

PR #519, opened 2026-06-25
chore(code-image): clarify mismatch / bad-image location messages

FileSystemGuy:
## Summary
- The hash-mismatch branch in `capture_or_verify_code_image` logged a bare `code image at: <path>` line after the primary spec error, leaving users to infer the path's relevance. Reworded to `the running code does not match the captured code image at: <path>`.
- The sibling missing/malformed-hash-file branch had the same problem; reworded to `affected code image at: <path>` to pair clearly with the preceding exception text.
- Literal VALR-02 / VALR-04 spec strings (`changes to the codebase are not allowed in a CLOSED run`, `all runs of this type must use the same codebase`) are untouched — only the supplementary location line is reworded.
- Pinned test assertions in `test_cli_code_image.py` updated to the new mismatch wording.

## Test plan
- [x] `pytest mlpstorage_py/tests/test_cli_code_image.py` — all 32 tests pass locally

### DIFF
--- a/mlpstorage_py/submission_checker/tools/code_image.py
+++ b/mlpstorage_py/submission_checker/tools/code_image.py
@@ -717,7 +717,7 @@ def capture_or_verify_code_image(args, env, log):
         matched = verify_source_against_image(source_root, code_dir, log)
     except (MissingHashFile, MalformedHashFile) as e:
         log.error(str(e))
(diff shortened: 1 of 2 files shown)


dslik (approved):
This is clearer. Thanks.

PR #524, opened 2026-06-25
Fix #521: kvcache fails fast when --results-dir is not on a shared filesystem

### LINKED ISSUE #521: KV Cache Benchmark: Result files scattered across nodes, aggregation fails without shared filesystem
When running the KV Cache benchmark across multiple nodes via MPI, each rank writes its result JSON file to the local filesystem. However, the final result aggregation (_aggregate_option_results() in kvcache.py) runs only on the
  controller node and uses glob.glob() to find kvcache_results_*.json files locally.

  This means that without a shared filesystem mounted on all nodes, the controller node can only find its own rank's results, and all remote ranks' result files are invisible to the aggregation step.

  Steps to Reproduce:

  1. Run the KV Cache benchmark across multiple hosts without a shared filesystem:
  mlpstorage closed kvcache run \
    --hosts host1 host2 host3 ... host16 \
    --num-processes 64 \
    --cache-dir /local/path/kvcache \
    --results-dir /local/path/results
  2. The mlperf_wrapper.py on each remote node writes kvcache_results_*.json to its local --results-dir
  3. Back on the controller node, _aggregate_option_results() globs for those files — but they don't exist locally

  Expected Behavior:

  The benchmark should either:
  - Option A: Collect result files from remote nodes after MPI execution completes (e.g., via mpirun file gathering or explicit SCP/rsync)
  - Option B: Use mpirun --output-file or MPI I/O to direct all output to a single shared location
  - Option C: Require and document that a shared filesystem must be mounted on all nodes for multi-host runs

FileSystemGuy:
## Summary

- Adds a proactive shared-FS probe to the kvcache benchmark: before launching any workload, each unique host in `--hosts` drops a sentinel file inside `--results-dir` via a one-rank-per-host `mpirun`. The controller then verifies it can see every sentinel locally. If any are missing, the run fails fast with a diagnostic that names the shared-FS requirement, lists which hosts wrote, and points at the remediation.
- Aggregation logs the same hint as a fallback when `partial_failure=True` in a multi-host run, in case the probe is bypassed or misses an edge case (e.g. a partial mount on a subset of nodes).
- `kvcache run --help` now documents the shared-FS requirement explicitly.

## Why

Per #521, multi-host kvcache runs write `kvcache_results_*.json` on whichever node each rank lands on, but `_aggregate_option_results()` globs them locally on the controller. Without a filesystem mounted at the same path on every host in `--hosts`, remote-host result files are invisible to aggregation and the run silently records `partial_failure` after the option loop has already run for many minutes.

This is the "Option C" fix from the issue — document and detect the requirement, fail closed early — chosen because it is the cheapest path to making the failure mode actionable. A future PR can add Option A (gather remote rank dirs back to the controller) for clusters without shared storage.

## On meeting the shared-FS requirement

The shared filesystem used for `--results-dir` **must not be the system-under-test**. The kvcache benchmark measures the SUT under load; if the result-aggregation FS is the same storage that's being benchmarked, every per-rank JSON write (and the cluster-collector / metadata writes alongside them) lands on the SUT and perturbs the scores being measured. This requirement is therefore not a request for additional benchmark-grade infrastructure — a plain NFS export from the controller to every host in `--hosts`, mounted at the same path, is sufficient and is what most submitters already have in place for logs and configs.

The probe pattern follows the existing `mlpstorage_py/cluster_collector.py` approach of executing a small script via `mpirun -host h1:1,h2:1,... -n N --map-by node`. Sentinel files are tagged with `OMPI_COMM_WORLD_RANK` / `PMI_RANK` and the rank's resolved hostname so the failure message can name which hosts couldn't reach the directory.

The probe is a no-op for single-host and all-localhost runs (cannot exhibit the bug) and is skipped under `--what-if`.

## Test plan

- [x] `uv run pytest tests/unit` — 1604 passed, 4 skipped, 0 failed
- [x] `uv run pytest tests/unit/test_benchmarks_kvcache.py` — 87 passed (7 new in `TestProbeResultsDirShared`)
- [x] `uv run pytest tests/unit/test_cli_kvcache.py` — 60 passed (covers the updated help string parsing)
- [x] New test cases:
  - localhost-only short-circuits (no mpirun invoked)
  - single-host short-circuits
  - all sentinels present → returns cleanly
  - missing sentinels → raises `RuntimeError` with hostnames + remediation
  - `--hosts h1:4 h2:4` strips slots → probes 2 unique hosts with `:1`
  - `_execute_run` wires the probe in for multi-host
  - `--what-if` bypasses the probe

Closes #521

### DIFF
--- a/mlpstorage_py/benchmarks/kvcache.py
+++ b/mlpstorage_py/benchmarks/kvcache.py
@@ -17,13 +17,16 @@ Classes:
 
 import json
 import os
+import shlex
 import sys
 import time
+import uuid
 from pathlib import Path
 from typing import Dict, Any, List
 from statistics import fmean
 
 from mlpstorage_py.benchmarks.base import Benchmark
+from mlpstorage_py.cluster_collector import _is_localhost
 from mlpstorage_py.config import (
     BENCHMARK_TYPES,
     KVCACHE_DEFAULT_DURATION,
@@ -283,6 +286,13 @@ class KVCacheBenchmark(Benchmark):
             processes_per_node=npernode,
         )
 
+        # Issue #521: rank result JSONs are written on the node where each rank
+        # lands, but aggregation globs them locally on the controller. Without a
+        # shared filesystem, remote-host results are invisible and the run
+        # silently records partial_failure. Fail fast with an actionable error.
+        if not getattr(self.args, 'what_if', False):
+            self._probe_results_dir_shared(hosts)
+
         option_results = {}
         for option in [1, 2, 3]:
             option_kv_args = self._build_option_kvcache_args(option, is_closed)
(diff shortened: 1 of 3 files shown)


dslik (approved):
This is an important check. Manual inspection of code looks good. Will test merged version.

PR #525, opened 2026-06-25
chore: bump version 3.0.19 -> 3.0.20

FileSystemGuy:
## Summary
- Bumps `mlpstorage` version from `3.0.19` → `3.0.20` in `pyproject.toml`.
- Regenerates `uv.lock` (only the editable `mlpstorage` entry's version line moves; no dependency churn).

Kept as a **draft** so it can be promoted / merged once whatever follow-up content it ships alongside (e.g., a `dlio-benchmark` pin bump after mlcommons/DLIO_local_changes#29 lands, addressing the mlcommons/storage#487 reopen) is ready to land in the same release.

## Test plan
- [x] `git diff` shows only the two version-line bumps.
- [x] `uv lock` clean, no dependency churn (`Resolved 108 packages`, single `Updated mlpstorage` line).


russfellows (approved):
LGTM

Issue #447, opened 2026-06-15
Training crashes with SemLock._rebuild -> FileNotFoundError


rread:
### Summary

It appears DLIO is sharing fork-context semaphores with spawn-context  workers. 

I see SemLock file not found error when I run training benchmarks:

```
[OUTPUT] 2026-06-15T19:35:24.345631 Starting epoch 1: 10 steps expected
[OUTPUT] 2026-06-15T19:35:24.346073 Starting block 1
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.12/multiprocessing/spawn.py", line 122, in spawn_main
    exitcode = _main(fd, parent_sentinel)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/multiprocessing/spawn.py", line 132, in _main
    self = reduction.pickle.load(from_parent)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/multiprocessing/synchronize.py", line 115, in __setstate__
    self._semlock = _multiprocessing.SemLock._rebuild(*state)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory
```

### Workaround
Changing the DLIO config to use `fork` instead of `spawn` resolves this issue.  

```
diff --git a/configs/dlio/workload/retinanet_b200.yaml b/configs/dlio/workload/retinanet_b200.yaml
index 94b456d..11b2f16 100644
--- a/configs/dlio/workload/retinanet_b200.yaml
+++ b/configs/dlio/workload/retinanet_b200.yaml
@@ -20,6 +20,7 @@ reader:
   data_loader: pytorch
   batch_size: 24
   read_threads: 8
+  multiprocessing_context: fork
   file_shuffle: seed
   sample_shuffle: seed
```
### Reproduction
Without the default config, I see this: 

```
(storage) mlps@node18:~/storage$ ./mlpstorage open training retinanet datagen file \
    --data-dir /mnt/lustrefs/mlps/training \
    -p dataset.num_files_train=240 \
    --results-dir ~/results \
    -np 8 \
    --hosts $(hostname)
2026-06-15 19:34:41|STATUS: 
2026-06-15 19:34:41|STATUS: --- Run Configuration (mlpstorage 3.0.9) ---
[...]

PR #32, opened 2023-06-12
fix workload path in README

jeromecst:
fix the workload path in README

PR #29, opened 2023-05-27
Update README.md

TheKanter:
Fixing superscripts and phrasing

PR #533, opened 2026-06-26
chore: bump dlio-benchmark pin + version 3.0.20 -> 3.0.21

### LINKED ISSUE #466: Data-Loader OOM Issues
# Dataset Sharding Analysis: s3dlio + dlio_benchmark

**Date**: June 17, 2026  
**Scope**: Investigation only — no code changes proposed here

---

## Issues identified by @wolfgang-desalvador 

@wolfgang-desalvador @keithpij and @zhenghh04 please review. 

## Problem Statement

For workloads with tens of millions of files or objects, Python OOM occurs because every rank
currently holds the entire file list in RAM. The file list is built at startup before any
training begins, and every rank builds the same global list.

**Specific RAM cost** (as a concrete anchor):

| Dataset size | Path string (~60 bytes avg) | Python list overhead (~56 bytes/item) |
|-------------|----------------------------|--------------------------------------|
| 1 M files | ~60 MB | ~116 MB total |
| 10 M files | ~600 MB | ~1.16 GB per rank |
| 64 M files | ~3.84 GB | ~7.4 GB **per rank** |

With 8 ranks per node, 64 M files → ~60 GB RAM consumed just for file lists on a single
8-rank node, before a single byte of training data is read.

---

### dlio_benchmark: the problem is in `main.py` + `config.py`

**Root cause**: `main.py` lines 213-263 — every rank calls `storage.walk_node()` which
calls `list_objects()` and accumulates the **full global list** in `fullpaths`. This is then
stored in `ConfigArguments.file_list_train` (a `ClassVar[List[str]]`) — a process-global
singleton that all reader threads inherit.

```
main.py:walk_node() → storage.list_objects() → fullpaths (full global list, all ranks)
    ↓
ConfigArguments.file_list_train = fullpaths   ← ENTIRE list, same on every rank
    ↓
config.py:build_sample_map_iter()  ← slices the list by rank AFTER loading it all
config.py:get_global_map_index()   ← builds VirtualIndexMap from full list
    ↓
reader_handler.py:self._file_list = args.file_list_train  ← each reader gets full list
```

The rank-level sharding only happens **after** the full list is in RAM. For 10 M files,
every rank first allocates ~1 GB just to keep the list, then discards 7/8 of it.

### s3dlio `ParquetRowGroupDataset`: same pattern at a smaller scale

In `parquet_rg.rs::new()` (lines 180-188):
```rust
let file_uris: Vec<String> = run_on_global_rt(async move {
    store.list(&uri_prefix_owned, true).await
})
.filter(|u| u.ends_with(".parquet"))
.collect();
```

The full list is collected into `Vec<String>` before any construction proceeds. For
10 M Parquet files this is ~600 MB just for the URI strings in Rust. Python inherits this
through `file_uris_arc()`.

---

### dlio_benchmark — MPI-level sharding

`config.py:build_sample_map_iter()` shards at the **sample** level across MPI ranks after
the full list is loaded:

```python
samples_per_proc = int(math.ceil(total_samples / self.comm_size))
start_sample_index = samples_per_proc * self.my_rank
end_sample_index = samples_per_proc * (self.my_rank + 1) - 1
```

This correctly distributes work but does not reduce RAM — every rank still holds the full
list.

### dlio_benchmark — PyTorch worker sharding (TorchIterableDataset)

`torch_data_loader.py:TorchIterableDataset.__iter__()` already shards files across
PyTorch DataLoader workers:

```python
all_files = all_files[worker_info.id::worker_info.num_workers]
```

This is correct intra-rank sharding but again operates on a list that was already fully
loaded.

### s3dlio — `PyDataset.from_uris()`

`python_aiml_api.rs` already exposes:

```python
ds = s3dlio.PyDataset.from_uris(uris)
```

This accepts a pre-built URI list, so a caller can compute its shard and pass only its
portion. This is the right hook — but the caller must first enumerate the full list to slice
it, which defeats the purpose.

---

### Option A — Prefix-Based Deterministic Sharding (Recommended First Step)

**How it works**: Assign each rank a slice of the key space by deterministic prefix or
last-character hash. Storage APIs accept any prefix, so this costs only 1/N of the listing
work and RAM:

```
Rank 0: list("s3://bucket/train/000*/") + "001*/" + ... up to 1/N of prefix space
Rank 1: list("s3://bucket/train/032*/") + ...
```

For uniformly distributed keys (UUIDs, hash-named files, most training sets), this splits
perfectly. For sequential names (`sample_000000.parquet`, `sample_000001.parquet`), you can
shard by prefix range:

```
Rank 0: list prefix "sample_0" through "sample_1"
Rank k: list prefix covering the k-th bucket of the key range
```

**What needs to change**:
1. In s3dlio `list()` (Rust): add `start_after: Option<String>` and `end_before: Option<String>` parameters. AWS S3 `list_objects_v2` supports `StartAfter` natively. Azure Blob and GCS have equivalent parameters. This is a small, safe, additive change.
2. In dlio_benchmark `main.py`: compute the rank's prefix range, list only that range, never call `walk_node()` for the full prefix.

**Tradeoffs**:
- ✅ RAM: O(N/ranks) instead of O(N) — 8 ranks → 8× less RAM
- ✅ Listing latency: 8 parallel listings of 1/8 each → same wall-clock time
- ✅ No MPI barrier needed for listing — ranks list in parallel
- ⚠ Requires approximately uniform key distribution. Skewed datasets need a sampling step.
- ❌ Does not support global shuffle across all N files between epochs (see Option D).

### Option B — Manifest File / Pre-Built Shard Files

Generate one manifest per rank at data-generation time (or as a preprocessing step):

```
s3://bucket/manifests/train_rank0.txt   # 1/8 of file list, one URI per line
s3://bucket/manifests/train_rank1.txt
...
```

Each rank reads only its own manifest (tiny text file, ~60 MB for 1 M entries). This is
how large-scale PyTorch training pipelines (WebDataset, NVIDIA DALI) handle this at the
1 B+ file scale.

**What needs to change**:
1. dlio_benchmark: add a `manifest_file` config option; replace `walk_node()` with
   manifest read when set.
2. Optionally: add a `dlio_benchmark generate_manifest` sub-command.
3. s3dlio: no changes needed — `PyDataset.from_uris()` already accepts pre-built lists.

**Tradeoffs**:
- ✅ Zero listing overhead at training time
- ✅ Arbitrary shard sizes — manifests can be unequal for heterogeneous clusters
- ✅ Manifests can be pre-shuffled per epoch
- ⚠ Requires a manifest generation step before training (one-time cost or re-run for new shuffles)
- ❌ Re-shuffling between epochs requires generating new manifests or re-sorting at epoch start

### Option C — Lazy Streaming Listing (Avoids RAM, Not Sharding)

Instead of `list() → Vec<String>`, stream listing results through an iterator that yields
URIs on demand. Each worker pulls the next batch of URIs as it finishes previous work. This
eliminates the RAM spike but does not reduce total listing volume:

**s3dlio**: `list_objects_stream()` already exists in `s3_utils.rs` (line 760). It is a
Tokio async stream that pages through listing results incrementally. The Python API does
not currently expose a streaming listing interface.

**dlio_benchmark** `ParquetRowGroupDataset`: the `build_extents()` call in `parquet_rg.rs`
already streams concurrent footer fetches — but it first collects the full URI list.

**What needs to change**:
1. s3dlio: expose `list_stream(uri_prefix) → Iterator[str]` in the Python API. This is a
   wrapper around the existing `list_objects_stream()` Tokio stream, bridged to Python via
   `blocking_recv()`.
2. `ParquetRowGroupDataset::new()`: accept either `uri_prefix: &str` (current) or
   `file_uris_iter: impl Iterator<Item=String>` so callers can provide a streaming source.
3. dlio_benchmark: replace `walk_node()` with `s3dlio.list_stream()`.

**Tradeoffs**:
- ✅ O(1) RAM for listing — URIs never all in RAM simultaneously
- ✅ Works with any key distribution, no pre-processing needed
- ✅ Compatible with existing shuffle (shuffle the stream in a sliding window)
- ⚠ Inter-rank coordination: all ranks stream the same prefix, doing 8× the listing work
- ⚠ Shuffling a streaming list requires reservoir sampling or a partial buffer

### Option D — Coordinated Distributed Listing (Scalable + Shuffle-Safe)

One rank (rank 0) lists the full dataset, then broadcasts a shuffled permutation array
using only indices (not strings) to all other ranks. Each rank then lists its own shard
by index from the permutation.

This is how `dlio_benchmark` already works for sample-level shuffle
(`config.py:get_global_map_index()`), but `VirtualIndexMap` still requires the full
string list to exist somewhere.

**Improved version**:
1. Rank 0 only lists once and stores key hash digests (8 bytes each, not 60+ byte strings).
2. Rank 0 broadcasts a shuffled permutation of **indices** to all ranks.
3. Each rank reconstructs only its shard of string URIs (fetching names by range from storage).

At 64 M files: index array = 512 MB (int64) on rank 0 vs 4 GB string list. After broadcast,
each of 8 ranks holds only 64 MB of indices, then fetches its ~8 M file names.

**What needs to change**:
1. s3dlio: expose a `list_keys_hashed(prefix) → List[u64]` or `list_count(prefix) → int`
   plus `list_range(prefix, start_after, limit) → List[str]` API.
2. dlio_benchmark: replace `walk_node()` with MPI-coordinated version above.
3. Requires MPI barrier at listing time (rank 0 lists first, then broadcasts).

**Tradeoffs**:
- ✅ Full global shuffle preserved
- ✅ Scales to 100 M+ files
- ⚠ More complex; needs MPI coordination
- ⚠ Rank 0 still holds the full index array (just 8 bytes/file vs 60+ bytes/file)

---

## Shuffle Between Epochs — Analysis

This is the hardest constraint. Current dlio_benchmark shuffle (`config.py:reconfigure()`):

```python
np.random.shuffle(self.file_list_train)   # in-place shuffle of full list
```

This requires the full list in RAM on every rank. Options:

| Approach | Epoch shuffle support | RAM after listing |
|----------|----------------------|-------------------|
| A (prefix shard) | ❌ No global shuffle (only local shard shuffle) | O(N/ranks) |
| B (manifest) | ✅ Pre-generate shuffled manifests per epoch | O(N/ranks) |
| C (streaming) | ⚠ Reservoir sample shuffle in sliding window | O(window_size) |
| D (distributed) | ✅ Full global shuffle via index broadcast | O(N×8B / ranks) |

**Practical recommendation for shuffle**: For true global shuffle at scale, pre-generate
per-epoch manifests (Option B). The manifest for epoch k is the epoch k-1 manifest
reshuffled deterministically (seeded by epoch number). A manifest of 10 M entries is only
~600 MB of compressed text — trivial to store in object storage.

---

### Immediate (dlio_benchmark only, no s3dlio changes needed)

1. **Add `manifest_file` config option** to `main.py` — when set, skip `walk_node()` and
   read the rank's manifest file instead. This requires ~20 lines of code change in
   `main.py` and `config.py`. Zero risk to existing functionality (opt-in config).

2. **Add `dlio generate_manifest` sub-command** — lists the full dataset once (can be run
   from rank 0 only), shards into N files, optionally shuffles, writes to storage.

### Short term (s3dlio changes, small)

3. **Expose streaming listing in Python API** — wrap existing `list_objects_stream()`
   as `s3dlio.list_stream(uri, opts) → iterator`. This is ~50 lines of Rust + PyO3 bridging
   the existing async stream to a Python iterator via `blocking_recv()`.

4. **Add `start_after` / `max_keys` to `store.list()`** — the AWS SDK `StartAfter` field
   is already threaded through the `list_objects_v2` builder in `s3_utils.rs`. Exposing it
   in the `ObjectStore` trait and Python API enables prefix-range sharding.

### Medium term

5. **`ParquetRowGroupDataset` accept pre-built URI list** — add a second constructor:
   `ParquetRowGroupDataset::from_uris(uris: Vec<String>, ...)`. This allows the caller to
   shard however it wants (manifest, prefix range, MPI shard) and pass only the shard's
   URIs to the DataLoader. The DataLoader then never sees the full list.

6. **`s3dlio.PyDataset.from_uris()` + streaming listing in dlio** — use the already-existing
   `from_uris()` API with a streaming listing call to drip URIs in batches, capping peak RAM
   at `batch_size × avg_uri_length`.

---

## Summary Table

| Option | dlio change complexity | s3dlio change complexity | RAM savings | Shuffle support | Recommended? |
|--------|----------------------|--------------------------|-------------|-----------------|--------------|
| A: Prefix shard | Medium | Small | O(N/ranks) | Local only | ✅ Good starting point |
| B: Manifest files | Small | None | O(N/ranks) | Full (pre-computed) | ✅ **Best for production** |
| C: Streaming list | Medium | Small | O(1) | Reservoir only | ✅ Good for listing RAM |
| D: Distributed index | Large | Medium | O(N×8B/ranks) | Full | For 100 M+ at scale |

**Recommended path**: Implement Option B (manifests) first — it solves the OOM problem
immediately with minimal code risk and works with every existing reader and data loader. Then
implement Option C (streaming listing) in s3dlio to eliminate listing RAM entirely for new
code paths. Options A and D are best saved for when global shuffle at the 100 M+ file scale
becomes a requirement.

### LINKED ISSUE #472: infinite s3dlio list during training (retinanet)
Issue description:  
We have 4 clients with 755GB memory per client. This results in 50 million .3MiB Jpg files to be ingested to read back for training. s3dlio lists back the bucket and it has taken more than 12 hours and it is still listing.

1. Datasize:
mlpstorage closed training retinanet datasize --client-host-memory-in-gb 754 --max-accelerators 16 --accelerator-type b200 --num-client-hosts 4

gives:

RESULT: Number of training files: 50136788

2. Data generation:

mlpstorage closed training retinanet datagen object \
  --num-processes 64 \
  --hosts gizmoc5,gizmoc6,gizmoc7,gizmoc8 \
  --data-dir retinanet_64p \
  --params dataset.num_files_train=50136788 \
  --params storage.storage_type=s3 \
  --params storage.storage_root=retinanet1 \
  --mpi-params "-x AWS_ACCESS_KEY_ID -x AWS_SECRET_ACCESS_KEY -x AWS_REGION -x S3_ENDPOINT_URIS -x S3DLIO_MAX_CONCURRENCY -x S3DLIO_MULTIPART_THRESHOLD -x S3DLIO_PART_SIZE -x S3DLIO_CONNECTION_TIMEOUT -x S3DLIO_READ_TIMEOUT -x AWS_S3_MULTIPART_THRESHOLD -x RUST_LOG --map-by slot --mca pml ob1 --mca btl tcp,self --mca btl_openib_allow_ib 0 --mca btl_openib_warn_no_device_params_found 0" \
  --allow-run-as-root[11:59 AM]export AWS_ACCESS_KEY_ID=user1

export AWS_SECRET_ACCESS_KEY=iOpHmc2afcbYuAfAapfWljfxtbHM6XYalMC2GwD6
export AWS_REGION=us-east-1
export S3_ENDPOINT_URIS= http://ip1:9020,http://ip2:9020,http://ip3:9020,http://ip4:9020
export S3DLIO_MAX_CONCURRENCY=32
export S3DLIO_MULTIPART_THRESHOLD=1073741824
export S3DLIO_PART_SIZE=268435456
export S3DLIO_CONNECTION_TIMEOUT=300
export S3DLIO_READ_TIMEOUT=300
export RUST_LOG=s3dlio=debug,aws_sdk_s3=debug

3. Training:

mlpstorage open training retinanet run object \
  --data-dir retinanet_64p \
  --results-dir /tmp/result.20260618.033618/mlperf/training/retinanet/run_20260618_033720 \
  --accelerator-type b200 \
  --num-accelerators 256 \
  --client-host-memory-in-gb 755 \
  --hosts gizmoc5,gizmoc6,gizmoc7,gizmoc8 \
  --params storage.storage_type=s3 \
  --params storage.storage_root=retinanet1 \
  --params dataset.num_files_train=50204282 \
  --mpi-params "-x AWS_ACCESS_KEY_ID -x AWS_SECRET_ACCESS_KEY -x AWS_REGION -x S3_ENDPOINT_URIS -x S3DLIO_MAX_CONCURRENCY -x S3DLIO_MULTIPART_THRESHOLD -x S3DLIO_PART_SIZE -x S3DLIO_CONNECTION_TIMEOUT -x S3DLIO_OPERATION_TIMEOUT_SECS -x S3DLIO_READ_TIMEOUT -x RUST_LOG --map-by slot --mca pml ob1 --mca btl tcp,self --mca btl_openib_allow_ib 0 --mca btl_openib_warn_no_device_params_found 0" \
  --allow-run-as-root

export AWS_ACCESS_KEY_ID=user1
export AWS_SECRET_ACCESS_KEY=iOpHmc2afcbYuAfAapfWljfxtbHM6XYalMC2GwD6
export AWS_REGION=us-east-1
export S3_ENDPOINT_URIS=http://ip1:9020,http://ip2:9020,http://ip3:9020,http://ip4:9020
export S3DLIO_MAX_CONCURRENCY=32
export S3DLIO_MULTIPART_THRESHOLD=1073741824
export S3DLIO_PART_SIZE=268435456
export S3DLIO_CONNECTION_TIMEOUT=600
export S3DLIO_OPERATION_TIMEOUT_SECS=600
export S3DLIO_READ_TIMEOUT=600
export RUST_LOG=s3dlio=info,aws_sdk_s3=warn

### LINKED ISSUE #504: Datagen fails when using the s3torchconnector
The output I am seeing is almost identical to the output in issues 392 and 393. Basically, the connection to S3 via the s3torchconnector library is failing, so the jobs queue is filling up faster than it can be emptied, causing an out-of-memory condition. 

I think the solution should be twofold:

1. Correct the connection error. Each of the three libraries we support wants connection and bucket information in a slightly different shape. I suspect that this is the issue here. I think the environment variables are set up properly. 
2. If there is a connection problem, an error should be raised, and the datagen operation aborted. I believe we are trapping and continuing.

### LINKED ISSUE #528: Training crashes with `RuntimeError: could not unlink the shared memory file /torch_*`
> Related: #447 / #460 — same root cause (`systemd-logind RemoveIPC=yes`), **different shm vector**. PR #460 only mitigates the `SemLock` path; the PyTorch tensor-storage path (`/dev/shm/torch_*`) is still vulnerable even with `multiprocessing_context: fork`.

## Summary

After applying the workaround from #460 (`reader.multiprocessing_context: fork` in the workload YAML) and confirming via DLIO config dump that `"multiprocessing_context": "fork"` is in effect, multi-node training runs of `unet3d` still crash mid-epoch with:

```
RuntimeError: could not unlink the shared memory file /torch_<pid>_<id>_<n> : No such file or directory (2)
```

This is a **distinct failure mode** from #447. PR #460 fixes the `multiprocessing.SemLock._rebuild -> FileNotFoundError` symptom but does not address the PyTorch named tensor-storage shm (`/dev/shm/torch_*`), which is reaped by the same `systemd-logind RemoveIPC=yes` mechanism on Ubuntu 24.04.

## Environment

- mlpstorage 3.0.16
- PyTorch DataLoader path; `reader.read_threads=64`; `reader.multiprocessing_context=fork` (verified in `dlio_config/config.yaml` of the run results dir)
- Python 3.12; PyTorch default tensor sharing strategy = `file_system`
- HPC-X v2.25.1 OpenMPI `mpirun` via PRRTE `ssh` PLM (not PMIx)
- Ubuntu 24.04 LTS compute nodes; `/etc/systemd/logind.conf` default `RemoveIPC=yes`
- No `loginctl enable-linger` for the launching user
- Slurm; up to 12 compute nodes used

## Symptom (with `multiprocessing_context: fork`)

 `unet3d` `closed training run`. Excerpt from worker traceback in `mlperf_storage.log`:

```
File ".../dlio_benchmark/data_loader/torch_data_loader.py", line 671, in next
  ...
RuntimeError: Caught RuntimeError in DataLoader worker process 40.
Original Traceback (most recent call last):
  ...
RuntimeError: could not unlink the shared memory file /torch_29215_1991744185_2 : No such file or directory (2)
```

And from the metadata JSON written by mlpstorage for the same run:

```json
"reader": {
  ...
  "multiprocessing_context": "fork",
  ...
}
```

So we are unambiguously on the fork path, with PR #460's mitigation active.

## Reproduction

1. Ubuntu 24.04 compute nodes, default `/etc/systemd/logind.conf` (`RemoveIPC=yes`), launching user has **no** linger enabled (`loginctl show-user $USER --property=Linger` → `Linger=no`).
2. Multi-node MPI launch via `mpirun`
3. Workload YAML pins `reader.multiprocessing_context: fork` (the #460 fix) and `reader.read_threads >= 8` (so PyTorch DataLoader workers exist and use named tensor-storage shm for IPC).
4. Run `closed training unet3d run file` at ≥ 4 nodes.

Failure is probabilistic per worker process; per-node probability is roughly the fraction of the run during which the user is sessionless. Aggregate failure rate scales as `1 - (1-p)^N`. Observed empirically:

| Nodes | Outcome |
|---|---|
| 2 | usually succeeds |
| 4 / 6 / 8 | mixed — fails most of the time |
| 12 | fails every time observed |

## Root cause

There are **two distinct user-owned POSIX named-shm vectors** that `systemd-logind` with `RemoveIPC=yes` will reap when the user's login-session count transitions to 0:

| # | Pathname | Created by | Lifecycle | Covered by #460? |
|---|---|---|---|---|
| 1 | `/dev/shm/sem.mp-XXXXXXXX` | Python `multiprocessing.SemLock` | re-opened by name in spawn-context children → `FileNotFoundError: SemLock._rebuild` | ✅ Yes (fork avoids re-open by name) |
| 2 | `/dev/shm/torch_<pid>_<id>_<n>` | `torch.Tensor.share_memory_()` / DataLoader IPC (default sharing strategy = `file_system`) | worker creates, main maps, **worker `shm_unlink()`s by name** after main has it mapped → `RuntimeError: could not unlink the shared memory file ... No such file or directory (2)` | ❌ **No** |

PR #460 closes vector #1 but vector #2 is independent: switching the multiprocessing start method from `spawn` to `fork` has no effect on PyTorch's named tensor-storage shm, because PyTorch's `THManagedMapAllocator` uses `shm_open("/torch_*")` and a subsequent `shm_unlink()` by name regardless of fork vs spawn. If `systemd-logind` reaps `/dev/shm/torch_*` between create and unlink, the unlink call returns `ENOENT` and PyTorch raises.

The trigger is unchanged from #447: PRRTE `ssh` bootstrap connections exit immediately after their child `prted` is reparented under `init` on each compute node. While the rank processes are still very much alive (parented under `prted`), `systemd-logind` observes the user's session count drop to 0 and fires its IPC sweep, unlinking *everything* user-owned in `/dev/shm/`.

## Why the workaround in #460 is insufficient

The advisory added in #460 (`environment/systemd_ipc.py` detecting `RemoveIPC=yes` and recommending `loginctl enable-linger`) is the *correct* mitigation. But the YAML change (pinning `multiprocessing_context: fork`) only addresses the SemLock symptom; it does not protect PyTorch's named tensor-storage shm, so customers who apply *only* the YAML half of the fix (e.g. an in-place patch of an installed `mlpstorage` without OS-level changes) still hit a crash that looks "the same" but isn't.

FileSystemGuy:
Closes #466
Closes #472
Closes #504
Closes #528

## Summary

- `pyproject.toml`: `3.0.20` → `3.0.21`.
- `uv.lock`: regenerated via `uv lock --upgrade-package dlio-benchmark`; pin moves `ae1247a6` (DLIO PR #29 merge) → `75d129b1` (current DLIO main HEAD).

## What's in the DLIO bump

Five DLIO PRs have landed since the previous pin and are pulled in here:

| DLIO PR | mlcommons/storage issue | What it does |
|---|---|---|
| [#30](https://github.com/mlcommons/DLIO_local_changes/pull/30) | #504 | Propagate `dataset.skip_listing` through `LoadConfig` so the CLI / YAML override actually reaches the listing layer. |
| [#31](https://github.com/mlcommons/DLIO_local_changes/pull/31) | #504 | `data_generator` fast-fail in the async upload pipeline. On the first `put_data()` exception, stop submitting new uploads and re-raise promptly. Bounds `_futures` growth to `n_workers` after the first failure — kills the unbounded-list OOM regardless of trigger. |
| [#32](https://github.com/mlcommons/DLIO_local_changes/pull/32) | #472 | @russfellows' preflight in `ObjStoreLibStorage`: credentials check, s3torchconnector silent-route-to-AWS guard, per-library bucket reachability probe. |
| [#34](https://github.com/mlcommons/DLIO_local_changes/pull/34) | #528 | Wrap PyTorch shm-reap `RuntimeError` with an actionable banner pointing operators at `loginctl enable-linger` (primary fix) and `DLIO_TORCH_SHARING_STRATEGY=file_descriptor` + `ulimit -n` raise (fallback). Adds the `DLIO_TORCH_SHARING_STRATEGY` env-var hook so the fallback is actually usable. |
| [#35](https://github.com/mlcommons/DLIO_local_changes/pull/35) | #466 | Eliminate double-allocation OOM in flat-layout file listing. |

This release effectively resolves both directions of the storage#504 OOM (preflight surfaces config errors clearly; fast-fail bounds the OOM regardless of cause), wires up the operator-facing banner for storage#528, fixes skip_listing propagation for storage#504, and closes #466's flat-layout listing OOM.

## Test plan

- [x] `git diff` shows only the two pyproject.toml / uv.lock version lines + the dlio-benchmark SHA in uv.lock.
- [x] `uv lock --upgrade-package dlio-benchmark` clean — no dependency churn beyond dlio-benchmark and the editable mlpstorage entry.
- [ ] No mlpstorage code change in this PR — behavioral testing for the five DLIO PRs lives in their respective PRs / issues. Worth a smoke run on a real object-storage workload before merging, since #32 changes the `ObjStoreLibStorage.__init__` path and #31 changes the upload-failure path.

## Coordination

- Companion PR **mlcommons/storage#532** (still draft) updates the `systemd_ipc.py` startup advisory text to reference both #447 and #528 vectors and the new `DLIO_TORCH_SHARING_STRATEGY` fallback — pairs with DLIO #34 in this bump. Can be promoted out of draft once the team has eyes on this PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "mlpstorage"
-version = "3.0.20"
+version = "3.0.21"
 description = "MLPerf Storage Benchmark Suite"
 readme = "README.md"
 license = {text = "Apache-2.0"}
--- a/uv.lock
+++ b/uv.lock
@@ -237,7 +237,7 @@ wheels = [
 [[package]]
 name = "dlio-benchmark"
 version = "3.0.2"
-source = { git = "https://github.com/mlcommons/DLIO_local_changes.git?branch=main#ae1247a6d430f50acc8552856462d75dd88b8b52" }
+source = { git = "https://github.com/mlcommons/DLIO_local_changes.git?branch=main#b63fef4102cce46432e8f6a1ba6b655f79ff70f5" }
 dependencies = [
     { name = "dgen-py", marker = "sys_platform == 'linux'" },
     { name = "h5py", marker = "sys_platform == 'linux'" },
@@ -518,7 +518,7 @@ wheels = [
 
 [[package]]
 name = "mlpstorage"
-version = "3.0.20"
+version = "3.0.21"
 source = { editable = "." }
 dependencies = [
     { name = "dlio-benchmark", marker = "sys_platform == 'linux'" },
--- a/uv.lock
+++ b/uv.lock
@@ -237,7 +237,7 @@ wheels = [
 [[package]]
 name = "dlio-benchmark"
 version = "3.0.2"
-source = { git = "https://github.com/mlcommons/DLIO_local_changes.git?branch=main#b63fef4102cce46432e8f6a1ba6b655f79ff70f5" }
+source = { git = "https://github.com/mlcommons/DLIO_local_changes.git?branch=main#75d129b1ca3b659d2b5095a6fc320f4033a50b2e" }
 dependencies = [
     { name = "dgen-py", marker = "sys_platform == 'linux'" },
     { name = "h5py", marker = "sys_platform == 'linux'" },


FileSystemGuy:
@russfellows

russfellows (approved):

Issue #349, opened 2026-04-23
Running training with --open doesn't seem to be recognized


dslik:
When I try to run a training workload with `--open`, mlpstorage still says `Running the benchmark without verification for open or closed configurations. These results are not valid for submission. Use --open or --closed to specify a configuration.`

I also get a crash later on:

```
dslik@frosta:~/scratch/mlp3/storage$ ./mlpstorage training run --open --hosts 127.0.0.1 --num-client-hosts 1 --client-host-memory-in-gb 128 --num-accelerators 2 --accelerator-type b200 --model retinanet --param dataset.num_files_train=425000 --file --data-dir /home/dslik/scratch --results-dir /home/dslik/scratch --allow-run-as-root
Setting attr from num_accelerators to 2
Hosts is: ['127.0.0.1']
Hosts is: ['127.0.0.1']
⠋ Validating environment... 0:00:002026-04-22 23:24:12|INFO: Environment validation passed
2026-04-22 23:24:12|STATUS: Benchmark results directory: /home/dslik/scratch/training/retinanet/run/20260422_232412
2026-04-22 23:24:12|INFO: Created benchmark run: training_run_retinanet_20260422_232412
2026-04-22 23:24:12|STATUS: Verifying benchmark run for training_run_retinanet_20260422_232412
2026-04-22 23:24:12|RESULT: Minimum file count dictated by dataset size to memory size ratio.
2026-04-22 23:24:12|STATUS: Closed: [CLOSED] Closed parameter override allowed: dataset.num_files_train = 425000 (Parameter: Overrode Parameters)
2026-04-22 23:24:12|ERROR: INVALID: [INVALID] Insufficient number of training files (Parameter: dataset.num_files_train, Expected: >= 16550423, Actual: 425000)
2026-04-22 23:24:12|STATUS: Benchmark run is INVALID due to 1 issues ([RunID(program='training', command='run', model='retinanet', run_datetime='20260422_232412')])
2026-04-22 23:24:12|WARNING: Running the benchmark without verification for open or closed configurations. These results are not valid for submission. Use --open or --closed to specify a configuration.
```

snip

```
--------------------------------------------------------------------------
[OUTPUT] 2026-04-22T23:24:15.550253 Running DLIO [Training] with 2 process(es)
[WARNING] The amount of dataset is smaller than the host memory; data might be cached after the first epoch. Increase the size of 
dataset to eliminate the caching effect!!!
[...]

PR #558, opened 2026-06-26
test(cli): drop vestigial --open/--closed boolean-flag tests

FileSystemGuy:
## Summary
PR #412 reshaped the CLI from boolean \`--open\`/\`--closed\` flags into a positional \`mode\` argument (\`closed\`|\`open\`|\`whatif\`). Argparse no longer defines either flag — \`grep -rn 'add_argument.*--open|--closed' mlpstorage_py/\` returns zero hits — so:

- The 4 \`TestOpenClosedCLIFlags\` tests exercise nonexistent parser behavior. They were failing with SystemExit when argparse rejected the unknown flag.
- The 6 \`TestVerifyBenchmarkOpenFlag\` tests build pre-#412 \`args.Namespace\` shapes that \`verify_benchmark\` no longer consumes (it reads \`args.mode\` now). They were failing with \`AttributeError: 'Namespace' object has no attribute 'mode'\`.

All 10 tests aren't asserting any contract worth re-affirming on the current codebase. The post-#412 dispatch tests (\`TestVerifyBenchmarkPost412ModeDispatch\`) already cover the modern contract — that's the file's value going forward. Kept those plus the kvcache helper-predicate test.

Refreshed the module docstring to reflect the new scope and the history of why the file exists (links the original #349 → #412 → #352 thread for future archaeology).

## Test plan
- [x] \`uv run python -m pytest mlpstorage_py/tests/test_open_closed_flag_recognition.py -v\` → 5 passed (was 5 passed + 10 failed)
- [x] Verified the modern contract is still exercised by the kept \`TestVerifyBenchmarkPost412ModeDispatch\` class (4 tests) covering closed/open/whatif/closed-rejecting-OPEN-only-params dispatch.

### DIFF
--- a/mlpstorage_py/tests/test_open_closed_flag_recognition.py
+++ b/mlpstorage_py/tests/test_open_closed_flag_recognition.py
@@ -1,24 +1,28 @@
 #!/usr/bin/env python3
 """
-Regression tests for issue #349: ``--open`` was indistinguishable from "no flag".
+Regression tests for the closed/open/whatif mode dispatch in
+``Benchmark.verify_benchmark()``.
 
-Before the fix:
+History:
 
-* ``--open`` was defined with ``action="store_false", dest="closed"``, so it
-  produced the same Namespace (``args.closed=False``, no ``args.open``) as
-  passing nothing at all.
-* ``verify_benchmark()`` used ``hasattr(self.args, "open")`` which is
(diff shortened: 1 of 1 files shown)


russfellows (approved):
Sounds valid, assuming testing verified it.

PR #559, opened 2026-06-26
test(integration): init results-dir and pass --systemname in real_accumulation tests

FileSystemGuy:
## Summary
\`main._main_impl()\`'s orgname-resolution gate now requires every gated mode (closed/open/whatif) to have a \`mlperf-results.yaml\` sentinel pinned into the results-dir by \`mlpstorage init\`. The universal argparse plumbing additionally requires \`--systemname\` (or \`MLPSTORAGE_SYSTEMNAME\`) on emitting commands.

The three module fixtures in \`tests/integration/test_real_accumulation.py\` predated both, so every subprocess CLI invocation exited 2 with either \"\`--systemname/-sn\` is required\" or, for the vectordb/kvcache fixtures, \"[E101] results-dir has not been initialized\". All 12 tests in the file failed at fixture setup.

## Changes
- Define \`TEST_ORGNAME\` and \`TEST_SYSTEMNAME\` module-level constants and add a \`_canonical_prefix(results_dir, mode)\` helper that builds \`<results_dir>/<mode>/<orgname>/results/<systemname>/\` — one edit if the layout changes again.
- In each of the three module fixtures: run \`mlpstorage init <orgname> <results-dir>\` *before* the benchmark invocations, and add \`--systemname <test-sys>\` to every benchmark argv.
- Update every path-shape assertion to walk the canonical prefix instead of the (now incorrect) \`<results_dir>/<type>/...\` shape.
- For vector_database, also add the \`<index_type>\` segment (DISKANN — the default) between engine and command. Current production splits per index_type (Rules.md §2.1.27): on-disk shape is \`vector_database/<engine>/<index_type>/<command>/<datetime>/\`.
- Production now combines engine + index_type into the metadata's \`model\` slot (e.g. \`milvus_DISKANN\`) so the per-index_type workload grouping matches the per-index_type path split. Update the metadata-schema and discovery assertions to expect the combined token.
- The heterogeneous test now symlinks each per-fixture canonical-prefix's per-type subdir into the combined dir, so discovery walks exactly the trees the three fixtures produced.

## Test plan
- [x] 6/6 vectordb + kvcache tests pass locally:
  \`uv run python -m pytest tests/integration/test_real_accumulation.py -k 'vectordb or kvcache' -m '' -o addopts=\` → 6 passed
- [ ] The 5 training tests + heterogeneous test (which depends on the training fixture) need ~520GB free disk for unet3d's CAP-01 capacity gate; my dev box has ~133GB. The canonical-prefix change is the same mechanical pattern verified end-to-end on vectordb/kvcache, but a CI machine with sufficient disk should run them.

### DIFF
--- a/tests/integration/test_real_accumulation.py
+++ b/tests/integration/test_real_accumulation.py
@@ -31,6 +31,23 @@ from mlpstorage_py.rules.submission_checkers.training import (
 REPO_ROOT = Path(__file__).resolve().parents[2]
 MLPSTORAGE_CLI = REPO_ROOT / "mlpstorage"
 
+# Identity values pinned into each results-dir by `mlpstorage init` and
+# supplied to every benchmark CLI invocation. They surface as path segments
+# under the canonical layout:
+#   <results_dir>/<mode>/<TEST_ORGNAME>/results/<TEST_SYSTEMNAME>/<type>/...
+# Both are POSIX-filename-safe (Rules.md §2.1.1 ^[A-Za-z0-9._-]+$).
+TEST_ORGNAME = "TestOrg"
+TEST_SYSTEMNAME = "test-sys"
+
+
+def _canonical_prefix(results_dir: Path, mode: str = "whatif") -> Path:
+    """Return the canonical-layout prefix under ``results_dir`` produced by
+    a run of mode ``mode`` against this file's init/systemname identity.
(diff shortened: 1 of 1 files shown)


russfellows (approved):
Movin' on.

PR #561, opened 2026-06-26
fix(#543): coerce simple_bench CSV numeric columns; drop bad rows instead of crashing

### LINKED ISSUE #543: `mlpstorage closed vectordb run file` can fail after the timed VectorDB benchmark completes because `vdbbench.simple_bench.calculate_statistics()` crashes when reading per-process CSV files.
ChatGPT summary of issue:

========

The `query_iterator failed ... RESOURCE_EXHAUSTED` line is noisy, but in this run it successfully fell back to pk-cursor pagination, copied **1,000,000 / 1,000,000** vectors, built the FLAT index, and precomputed ground truth for **1000 queries**. That part worked. 

The actual bug is later: after all 16 benchmark processes ran, `simple_bench.py` crashed during final statistics calculation:

```text
TypeError: unsupported operand type(s) for +: 'float' and 'str'
...
File ".../simple_bench.py", line 1042, in calculate_statistics
    (all_data["timestamp"] + all_data["batch_time_seconds"]).max()
```

The wrapper then marked the run as failed:

```text
Benchmark failed after 1 iterations
```

Your follow-up CSV inspection strongly supports a real benchmark bug. Most `milvus_benchmark_p*.csv` files parse cleanly, but at least `milvus_benchmark_p2.csv` and `milvus_benchmark_p8.csv` have `batch_time_seconds` parsed as `str`, with a malformed row where `batch_time_seconds=True` and other fields are shifted/missing.  

So the issue is probably one of these:

1. `simple_bench.py` occasionally writes malformed rows to per-process CSV files.
2. `calculate_statistics()` assumes all CSV rows are valid numeric data and crashes instead of rejecting/reporting malformed rows.
3. A failed/partial row can invalidate the entire benchmark after the timed work has already completed.

The current upstream code path still appears vulnerable in statistics calculation: it concatenates the per-process CSVs and directly computes `all_data["timestamp"] + all_data["batch_time_seconds"]` without coercing or validating numeric columns first. ([[GitHub](https://raw.githubusercontent.com/mlcommons/storage/main/vdb_benchmark/vdbbench/simple_bench.py)][1])

## Suggested issue title

Use this:

**`vdb_benchmark simple_bench: malformed per-process CSV row causes TypeError during calculate_statistics`**

Alternative:

**`VectorDB run fails after benchmark execution when batch_time_seconds is parsed as string`**

## GitHub-ready issue summary

````markdown
## Summary

`mlpstorage closed vectordb run file` can fail after the timed VectorDB benchmark completes because `vdbbench.simple_bench.calculate_statistics()` crashes when reading per-process CSV files.

In my run, Milvus loaded correctly, the source collection had 1,000,000 vectors, the FLAT ground-truth collection was created successfully with 1,000,000 vectors, and ground truth was precomputed for 1000 queries. The 16 benchmark worker processes also ran for about 60 seconds and wrote `milvus_benchmark_p*.csv` files.

However, final statistics calculation failed with:

```text
TypeError: unsupported operand type(s) for +: 'float' and 'str'
````

at:

```python
(all_data["timestamp"] + all_data["batch_time_seconds"]).max()
```

The wrapper then reported:

```text
Benchmark failed after 1 iterations
```

## Command

```bash
./mlpstorage closed vectordb run file \
  --host 127.0.0.1 \
  --port 19530 \
  --config default \
  --collection mlps_1m_1shards_1536dim_uniform \
  --vector-dim 1536 \
  --num-query-vectors 1000 \
  --num-query-processes 16 \
  --batch-size 100 \
  --vdb-index DISKANN \
  --results-dir /mnt/mlperf/$HOSTNAME/vdb_results \
  --debug
```

The effective vdbbench command was:

```bash
uv run vdbbench \
  --config /home/ir/mlperf/storage/configs/vectordbbench/default.yaml \
  --host 127.0.0.1 \
  --port 19530 \
  --collection-name mlps_1m_1shards_1536dim_uniform \
  --processes 16 \
  --batch-size 100 \
  --runtime 60 \
  --report-count 100 \
  --output-dir /mnt/mlperf/init-ptc-rb07-u28/vdb_results/closed/Everpure/vector_database/milvus/DISKANN/run/20260626_100617 \
  --vector-dim 1536 \
  --search-limit 10 \
  --search-ef 200 \
  --num-query-vectors 1000 \
  --json-output
```

## Good parts of the run

The source collection was valid:

```text
Collection Name: mlps_1m_1shards_1536dim_uniform
Vector Count: 1000000
Dimension: 1536
Index Types: DISKANN
Metric Types: COSINE
```

FLAT ground truth setup succeeded:

```text
Copied 1000000/1000000 vectors (100.0%)
FLAT collection 'mlps_1m_1shards_1536dim_uniform_flat_gt' ready with 1000000 vectors.
Ground truth pre-computation complete: 1000 queries in 15.65s
Ground truth ready: 1000 queries pre-computed.
```

## Failure

After benchmark execution finished, statistics calculation crashed:

```text
Calculating recall from captured ANN results...
Calculating benchmark statistics...
TypeError: unsupported operand type(s) for +: 'float' and 'str'
```

Traceback points to:

```python
(all_data["timestamp"] + all_data["batch_time_seconds"]).max()
```

## CSV diagnosis

I inspected the per-process CSV files with pandas. Most files had numeric dtypes, but two files had malformed rows.

Example from `milvus_benchmark_p2.csv`:

```text
dtypes:
process_id                float64
batch_id                    int64
timestamp                 float64
batch_size                float64
batch_time_seconds            str
avg_query_time_seconds    float64
success                    object

batch_time_seconds: bad_non_numeric_rows=1

 process_id  batch_id  timestamp  batch_size batch_time_seconds  avg_query_time_seconds success
        NaN       100    0.43958    0.004396               True                     NaN     NaN
```

Example from `milvus_benchmark_p8.csv`:

```text
batch_time_seconds: bad_non_numeric_rows=1

 process_id  batch_id  timestamp  batch_size batch_time_seconds  avg_query_time_seconds success
        NaN       100   0.372112    0.003721               True                     NaN     NaN
```

This looks like a malformed or shifted CSV row. The `True` value appears under `batch_time_seconds`, which causes pandas to treat the column as string/object and later causes the statistics calculation to crash.

## Expected behavior

The benchmark should either:

1. Never write malformed rows to `milvus_benchmark_p*.csv`, or
2. Validate/coerce CSV columns during `calculate_statistics()` and fail with a clear diagnostic identifying the bad file and row.

It should not crash with a pandas `TypeError` after benchmark execution has already completed.

## Actual behavior

The benchmark run completed enough to execute queries and generate ground truth, but final statistics failed and the wrapper marked the whole benchmark as failed.

## Suggested fixes

Possible fixes:

1. Ensure every `writer.writerow()` in `simple_bench.py` writes all fields:

   * `process_id`
   * `batch_id`
   * `timestamp`
   * `batch_size`
   * `batch_time_seconds`
   * `avg_query_time_seconds`
   * `success`

2. In `calculate_statistics()`, coerce and validate numeric columns before arithmetic:

```python
for col in ["timestamp", "batch_size", "batch_time_seconds", "avg_query_time_seconds"]:
    all_data[col] = pd.to_numeric(all_data[col], errors="coerce")

bad_rows = all_data[
    all_data["timestamp"].isna()
    | all_data["batch_size"].isna()
    | all_data["batch_time_seconds"].isna()
    | all_data["avg_query_time_seconds"].isna()
]

if not bad_rows.empty:
    raise ValueError(
        "Invalid rows found in benchmark CSV files. "
        "Please check per-process CSV output."
    )
```

3. Include the source CSV filename and row number in the error message so users can diagnose which worker produced malformed output.

```

## My recommendation

File it. The benchmark itself reached the important measurement stage, but the official result was lost because two malformed CSV rows poisoned the final pandas aggregation. That is issue-worthy, especially because users should not need to locally patch `simple_bench.py` to get a completed `statistics.json`.
```

[1]: https://raw.githubusercontent.com/mlcommons/storage/main/vdb_benchmark/vdbbench/simple_bench.py "raw.githubusercontent.com"

FileSystemGuy:
Fixes #543.

## Summary

`vdbbench.simple_bench.calculate_statistics()` was crashing **after** the timed VectorDB benchmark had already completed:

\`\`\`
TypeError: unsupported operand type(s) for +: 'float' and 'str'
    at (all_data["timestamp"] + all_data["batch_time_seconds"]).max()
\`\`\`

The reporter diagnosed the root cause: a small number of rows in two per-process CSVs (`milvus_benchmark_p2.csv`, `milvus_benchmark_p8.csv`) carried a `batch_time_seconds` value of the string `True` instead of a float duration. Once any value in a numeric column is non-numeric, pandas demotes the whole column to `object` dtype, and the arithmetic on `(timestamp + batch_time_seconds).max()` fails. The 60s benchmark completes, ground-truth runs, all 16 worker processes write their CSVs — then the wrapper marks the whole run as failed during the post-run statistics computation.

This PR hardens the reader so that result-data corruption in a handful of rows out of millions no longer destroys the whole run. **Writer-side root-causing** (why a row occasionally has shifted/missing fields) is left as a follow-up — the reader-side defense is enough to recover the benchmark result so long as the malformed rows are a small fraction.

## Fix

In `vdb_benchmark/vdbbench/simple_bench.py::calculate_statistics`:

1. Tag every loaded row with its source CSV name so the diagnostic identifies affected workers.
2. Coerce `timestamp`, `batch_size`, `batch_time_seconds`, `avg_query_time_seconds` with `pd.to_numeric(..., errors="coerce")` before any arithmetic.
3. Identify rows where any required numeric column became NaN; log a clear warning naming the dropped count and affected file(s); drop those rows.
4. Continue with the remaining good rows. If every row was bad, fall through to the existing `"No valid data found"` error rather than raising.

The diagnostic line, when triggered, looks like:

\`\`\`
Warning: dropping 2 row(s) with non-numeric values in required columns.
Affected file(s): milvus_benchmark_p2.csv, milvus_benchmark_p8.csv.
See mlcommons/storage#543.
\`\`\`

## Tests

New file: `vdb_benchmark/tests/tests/test_issue_543_csv_coercion.py` — 8 tests:

| Test | What it covers |
|---|---|
| `test_calculate_statistics_recovers_from_reporter_malformed_row` | Exact 16-worker scenario from the issue body — 14 clean CSVs, 2 with one malformed row each in the reporter's exact shape (`batch_time_seconds=True`, leading + trailing columns empty). Asserts no crash and stats returned. |
| `test_calculate_statistics_drops_rows_with_non_numeric_values[…]` × 5 | Parametrized — any one non-numeric value in any of the four numeric columns (`timestamp`, `batch_time_seconds`, `batch_size`, `avg_query_time_seconds`) must not crash. |
| `test_calculate_statistics_returns_error_when_every_row_is_malformed` | Edge case: all rows malformed returns the existing error key, not a raise. |
| `test_calculate_statistics_unchanged_for_clean_csvs` | Sanity: clean CSVs still produce a complete stats dict (happy-path regression check). |

## Test plan

- [x] \`pytest vdb_benchmark/tests/tests/test_issue_543_csv_coercion.py\` — 8/8 pass
- [x] \`pytest vdb_benchmark/tests/\` — 144/144 pass (no regressions in the wider vdb suite)

### DIFF
--- a/vdb_benchmark/tests/tests/test_issue_543_csv_coercion.py
+++ b/vdb_benchmark/tests/tests/test_issue_543_csv_coercion.py
@@ -1,0 +1,173 @@
+"""
+Issue #543: simple_bench.calculate_statistics() must not crash on malformed
+per-process CSV rows.
+
+Reporter scenario: 16 worker processes ran a 60s benchmark, finished writing
+their per-process ``milvus_benchmark_p*.csv`` files, and statistics
+calculation crashed with::
+
+    TypeError: unsupported operand type(s) for +: 'float' and 'str'
+        at (all_data["timestamp"] + all_data["batch_time_seconds"]).max()
+
+The root cause was that a small number of rows in two of the CSVs had a
+``batch_time_seconds`` value of the string ``True`` instead of a float
+duration — once any value in that column is non-numeric, pandas falls back
+to ``object`` dtype for the whole column and the arithmetic on line 1042
+explodes.
+
+The fix is reader-side defense in ``calculate_statistics()``: coerce the
+numeric columns with ``pd.to_numeric(..., errors='coerce')``, drop rows
+that fail to coerce, log a clear diagnostic naming the dropped count and
+the source files, and proceed with the valid rows. The benchmark run
+should succeed as long as enough good rows remain.
+"""
+
+from __future__ import annotations
+
+import csv
+import sys
+from pathlib import Path
+
+import pytest
+
+# Allow ``from vdbbench...`` from the source tree without an editable install.
+sys.path.insert(0, str(Path(__file__).resolve().parents[3] / "vdb_benchmark"))
+
+from vdbbench.simple_bench import calculate_statistics, csv_fields  # noqa: E402
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _write_csv(path: Path, rows):
+    """Write a per-process CSV using the production header order."""
+    path.parent.mkdir(parents=True, exist_ok=True)
+    with open(path, "w", encoding="utf-8", newline="") as fh:
+        writer = csv.DictWriter(fh, fieldnames=csv_fields)
+        writer.writeheader()
+        for row in rows:
+            writer.writerow(row)
+
+
+def _good_row(process_id, batch_id, timestamp, batch_size=100,
+              batch_time=0.5, avg_query=0.005, success=True):
+    return {
+        "process_id": process_id,
+        "batch_id": batch_id,
+        "timestamp": timestamp,
+        "batch_size": batch_size,
+        "batch_time_seconds": batch_time,
+        "avg_query_time_seconds": avg_query,
+        "success": success,
+    }
+
+
+# Exact shape the reporter observed in milvus_benchmark_p2.csv and p8.csv:
+# process_id empty, batch_id=100, the trailing two numeric fields and
+# 'success' empty, batch_time_seconds set to the string 'True'.
+_REPORTER_MALFORMED_ROW = {
+    "process_id": "",
+    "batch_id": 100,
+    "timestamp": 0.43958,
+    "batch_size": 0.004396,
+    "batch_time_seconds": True,
+    "avg_query_time_seconds": "",
+    "success": "",
+}
+
+
+# ---------------------------------------------------------------------------
+# Reproducer: the exact shape from issue #543
+# ---------------------------------------------------------------------------
+
+
+def test_calculate_statistics_recovers_from_reporter_malformed_row(tmp_path):
+    """The reporter's exact CSV shape must not crash and must return stats.
+
+    Two worker CSVs (p2 and p8) each contain one malformed row matching the
+    issue body, surrounded by good rows. The other workers' CSVs are clean.
+    """
+    # 14 clean CSVs
+    for pid in [0, 1, 3, 4, 5, 6, 7, 9, 10, 11, 12, 13, 14, 15]:
+        _write_csv(
+            tmp_path / f"milvus_benchmark_p{pid}.csv",
+            [_good_row(pid, b, b * 0.5) for b in range(1, 11)],
+        )
+
+    # 2 CSVs with one malformed row each (exact reporter shape)
+    for pid in [2, 8]:
+        rows = [_good_row(pid, b, b * 0.5) for b in range(1, 11)]
+        rows.insert(5, _REPORTER_MALFORMED_ROW)
+        _write_csv(tmp_path / f"milvus_benchmark_p{pid}.csv", rows)
+
+    # Must not raise
+    stats = calculate_statistics(str(tmp_path))
+
+    # No 'error' key — calculation succeeded
+    assert "error" not in stats, (
+        f"calculate_statistics returned an error: {stats!r}"
+    )
+
+    # Stats should reflect ~16 * 10 = 160 good rows (2 malformed dropped)
+    assert stats.get("total_queries", 0) > 0
+    assert stats.get("total_time_seconds", 0) > 0
+
+
+# ---------------------------------------------------------------------------
+# Per-malformation coverage — any one bad numeric field should not crash
+# ---------------------------------------------------------------------------
+
+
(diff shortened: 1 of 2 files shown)


idevasena (approved):
Changes look good. Thank you @FileSystemGuy

PR #549, opened 2026-06-26
Support HPE Cray PALS mpiexec launcher (ALCF Crux/Polaris/Aurora)

zhenghh04:
## Problem

`mlpstorage_py/utils.py::generate_mpi_prefix_cmd()` builds **OpenMPI-only** flags for *both* `mpirun` and `mpiexec`. On ALCF Cray EX systems (Crux/Polaris/Aurora) the launcher is **HPE Cray PALS `mpiexec`**, which rejects those flags, so `--mpi-bin mpiexec` produces an argv PALS cannot parse and every run fails to launch.

Reproduced on **Crux** (`/opt/cray/pals/1.7/bin/mpiexec` v1.7.1, `cray-mpich/9.0.1`, `cray-pals/1.7.1`) — the generated prefix is identical for both binaries:

```
[mpirun]  -> mpirun  -n 8 -host localhost:8 --npernode 8 --bind-to none --map-by socket
[mpiexec] -> mpiexec -n 8 -host localhost:8 --npernode 8 --bind-to none --map-by socket
```

PALS `mpiexec --help` accepts `-n/--np`, `--ppn`, `--cpu-bind`, `--depth`, `--hosts` (bare, comma-separated), `--hostfile` — but **not** `-host h:slots`, `--npernode`, `--bind-to`, `--map-by`, `--mca`, `--oversubscribe`, or `--allow-run-as-root`.

## Fix

Add a PALS-native branch for `mpiexec` in `generate_mpi_prefix_cmd()` that emits, e.g.:

```
mpiexec -n 8 --ppn 4 --hosts host1,host2 --cpu-bind none
```

- `--ppn` is taken from `processes_per_node`, or `ceil(num_processes / n_hosts)` when unset.
- `--hosts` is the bare, de-duplicated, comma-separated host list (no `:slots`).
- `--cpu-bind none` is injected as a safe default for an I/O benchmark, and suppressed if the user supplies `--cpu-bind` via `--mpi-params`.
- **`mpirun` behaviour is unchanged** (OpenMPI path untouched).

## Tests

`tests/unit/test_utils.py::TestGenerateMpiPrefixCmd` — strengthened the `mpiexec` test and added cases for: PALS flags present, OpenMPI flags absent, multi-host `--ppn`, and user `--cpu-bind` override. Existing `mpirun` tests act as a regression guard.

## Notes / follow-up

[...]

PR #564, opened 2026-06-27
PALS mpiexec for cluster collector + CAP-02 probe (ALCF; follow-up to #549)

zhenghh04:
## Follow-up to #549 — PALS `mpiexec` in the cluster collector & CAP-02 probe

#549 fixed `utils.generate_mpi_prefix_cmd()` (the DLIO/benchmark launcher). But `mlpstorage_py/cluster_collector.py` builds its **own** launcher command lines in two places that still emit **OpenMPI-only** flags regardless of `--mpi-bin`:

1. `MPIClusterCollector._generate_mpi_command` — the per-host system-info gather (populates the system description in result metadata).
2. `run_shared_fs_probe` — the **CAP-02** shared-filesystem probe.

On HPE Cray **PALS** `mpiexec` (ALCF Crux/Polaris/Aurora) these flags (`-host h:slots`, `--bind-to`, `--map-by`, `--tag-output`, `--allow-run-as-root`) are rejected.

### Impact

- **Single-host runs:** harmless — the collector catches the failure and falls back to local-only collection (observed on Crux: PALS prints its usage, then `INFO: Falling back to local-only collection`).
- **Multi-host runs:** broken — e.g. `llama3-70b/405b/1t` checkpointing (8/64/128 nodes) and multi-node training. The cross-host gather fails (only node 0 captured) and the CAP-02 shared-FS probe can't launch.

### Fix

Branch both builders on launcher family, mirroring #549. For `mpiexec`, emit a PALS-native command (one process per host):

```
mpiexec -n N --ppn 1 --hosts h1,h2 --cpu-bind none python3 <script> ...
```

`mpirun` behaviour is unchanged (still `-host h:1,…`, `--bind-to none`, `--map-by node`, `--tag-output`, and `--allow-run-as-root` when requested). The PALS probe path omits `--tag-output`, so rank-0's `__CAP02_RESULT_BEGIN__/END__` stdout markers are parsed verbatim — there is no tag prefix to strip.

### Tests

`tests/unit/test_shared_fs_probe.py::TestLauncherFlags` captures the probe's generated argv and asserts:
- `mpiexec` → `--ppn 1`, `--hosts h1,h2`, `--cpu-bind none`; none of `--map-by`/`--bind-to`/`--tag-output`/`-host`.
- `mpirun` → unchanged (regression guard).

Verified on Crux (PALS 1.7.1, `cray-mpich/9.0.1`).

### DIFF
--- a/mlpstorage_py/cluster_collector.py
+++ b/mlpstorage_py/cluster_collector.py
@@ -2975,21 +2975,22 @@ class MPIClusterCollector:
         # Build host string with 1 slot per host (we only need one process per node)
         host_slots = [f"{host}:1" for host in unique_hosts]
 
-        # Select MPI binary
-        if self.mpi_bin == MPIRUN:
-            mpi_executable = MPI_RUN_BIN
-        elif self.mpi_bin == MPIEXEC:
-            mpi_executable = MPI_EXEC_BIN
+        # Select MPI binary + launcher-family flags (one process per node).
+        if self.mpi_bin == MPIEXEC:
+            # HPE Cray PALS mpiexec (ALCF Crux/Polaris/Aurora): use --ppn and a
+            # bare comma-separated --hosts list with --cpu-bind. PALS rejects the
+            # OpenMPI flags (-host h:slots, --bind-to, --map-by,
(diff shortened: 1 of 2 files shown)


russfellows (approved):
Huihuo said its cool.

PR #574, opened 2026-06-29
fix(cap-01): skip capacity gate for object-storage destinations

gaikwadabhishek:
- CAP-01 ran statvfs on the destination → E401 "cannot determine free space" on s3:// targets
- _capacity_gate_destination() now returns None for object storage (s3/s3_torch)
- _pre_execution_gate then skips the gate via its existing remote-backend escape hatch
- local/direct_fs runs unchanged


gaikwadabhishek:
I have mentioned my github username in my MLCommons profile. Where can I access the CLA form?

gaikwadabhishek:
already fixed in main

PR #579, opened 2026-06-29
fix(#568): skip CAP-01 statvfs for object-mode DLIO runs

### LINKED ISSUE #568: CAP-01 capacity gate fails for training/checkpointing in `object` (S3) mode — `statvfs` runs on the `s3://` URI
## Summary
For `training` and `checkpointing` benchmarks in **`object` (S3) mode**, the
pre-execution CAP-01 capacity gate runs a local `os.statvfs()` against the
object-storage destination (an `s3://…` URI). The path never exists locally, so
the parent-walk exhausts and the run aborts before any work starts:

```
[E401] CAP-01: cannot determine free space — no valid parent for s3://mlc-s3/unet3d/data
  Details: Path: s3://mlc-s3/unet3d/data; Operation: cap01-check
```

The "remote backend" escape hatch (`_capacity_gate_destination()` returning
`None`, logged as *"CAP-01 skipped: destination not local"*) was wired for
**VectorDB and KVCache only** — not for training/checkpointing object mode.

## Environment
- `mlpstorage` version: **3.0.24**
- Storage library: `s3dlio` (`STORAGE_LIBRARY=s3dlio`), S3-compatible endpoint
- Mode: `closed`, both `training` (unet3d) and `checkpointing` (llama3-8b)
- `--data-dir s3://<bucket>/…` (training) / `--checkpoint-folder s3://<bucket>/…` (checkpointing)

## Steps to reproduce
```bash
# .env has valid AWS_ENDPOINT_URL / keys / BUCKET; bucket is reachable.
mlpstorage closed training unet3d datagen object \
  --hosts host0 host1 --num-processes 16 \
  --data-dir s3://mybucket/unet3d/data \
  --results-dir /shared/results --systemname sys \
  --allow-run-as-root \
  --params dataset.num_files_train=28000 dataset.num_subfolders_train=0
```
Result: aborts at CAP-01 with the `[E401] no valid parent for s3://…` error
above. Same failure for `checkpointing run object` (the destination becomes
`os.path.join(checkpoint_folder, model)`, e.g. `s3://mybucket/checkpoints/llama3-8b`).

## Expected
In object mode the local-disk capacity gate is meaningless and should be skipped
(as it already is for remote VectorDB/KVCache backends), allowing the run to
proceed. The injected storage params confirm the tool already knows it's object
mode:
```
--object mode: loaded credentials from .env ...
--object mode: injected storage params (storage_type=s3, storage_root=mybucket, library=s3dlio, uri_scheme=s3, force_path_style=True)
```

## Actual
The run aborts at the gate. `--skip-validation` does **not** help — CAP-01 lives
in `Benchmark._pre_execution_gate()`, which is independent of
`validate_benchmark_environment()` (the part `--skip-validation` gates).

## Root cause
`mlpstorage_py/benchmarks/dlio.py`:

- `TrainingBenchmark._capacity_gate_destination()` (~L761) returns
  `self.args.data_dir` unconditionally.
- `CheckpointingBenchmark._capacity_gate_destination()` (~L903) returns
  `os.path.join(self.args.checkpoint_folder, self.args.model)` unconditionally.

Neither checks for object mode, so `check_capacity_4field()`
(`mlpstorage_py/benchmarks/capacity_gate.py`) receives an `s3://` URI and does:

```python
check_path = destination_path
while not os.path.exists(check_path):          # never true for s3://…
    parent = os.path.dirname(check_path)
    if parent == check_path:
        raise FileSystemError("CAP-01: cannot determine free space — no valid parent …")
    check_path = parent
```

which always raises for a URI. The escape hatch already exists and is documented
in `base.py` (the "A8" path) — it's simply not used here.

## Proposed fix
Return `None` from `_capacity_gate_destination()` when the destination is an
object-storage URI (or, equivalently, when `data_access_protocol == 'object'` /
`storage.storage_type` is one of the object types), mirroring the VectorDB/KVCache
escape hatch. Sketch:

```python
def _capacity_gate_destination(self):
    # Object/S3 destinations are not local — skip the statvfs gate (A8 escape hatch).
    if getattr(self.args, 'data_access_protocol', None) == 'object':
        return None
    return self.args.data_dir            # (checkpointing: join(checkpoint_folder, model))
```

Alternatively, teach `check_capacity_4field()` to detect a URI scheme
(`urllib.parse.urlparse(...).scheme in {s3,s3a,az,gs}`) and no-op.

## Impact
Blocks **all** object-storage training and checkpointing runs. File/POSIX mode is
unaffected.

FileSystemGuy:
## Summary
- Fixes #568. In `object` (S3) mode, the CAP-01 pre-execution capacity gate ran `os.statvfs()` against the `s3://…` destination; the parent-walk in `check_capacity_4field()` exhausted and aborted every training/checkpointing run with `[E401] CAP-01: cannot determine free space — no valid parent for s3://…`. `--skip-validation` did not help — CAP-01 lives in `_pre_execution_gate()`, separate from `validate_benchmark_environment()`.
- `TrainingBenchmark._capacity_gate_destination()` and `CheckpointingBenchmark._capacity_gate_destination()` now return `None` when `data_access_protocol == 'object'`, firing the existing A8 remote-backend escape hatch in `Benchmark._pre_execution_gate` (the same hatch VectorDB/KVCache already use). Logged as `CAP-01 skipped: destination not local`.
- File/POSIX runs unaffected: missing `data_access_protocol` attribute falls through to the existing `statvfs` path.
- Drive-by: the dep-stub block in `tests/unit/test_capacity_gate.py` raised `ModuleNotFoundError` from `find_spec("pyarrow.ipc")` after the parent `pyarrow` was stubbed with `MagicMock`, blocking collection. Wrapped the lookup so it degrades to "treat as missing."

## Test plan
- [x] RED commit: 4 new tests in `tests/unit/test_capacity_gate.py` (`test_destination_is_none_in_object_mode` + `test_destination_*_when_protocol_attribute_missing` × 2 benchmarks). 2 new tests fail on RED with the exact `s3://…` URI return.
- [x] GREEN commit: all 4 suites pass on the branch.
  - `tests/`: 2385 passed
  - `mlpstorage_py/tests`: 780 passed, 1 xfailed
  - `vdb_benchmark/tests`: 144 passed
  - `kv_cache_benchmark/tests`: 238 passed
- [ ] Reporter (@jadnov) re-runs the original repro on this branch and confirms training/checkpointing object-mode now proceeds past the gate.

### DIFF
--- a/mlpstorage_py/benchmarks/dlio.py
+++ b/mlpstorage_py/benchmarks/dlio.py
@@ -3,6 +3,7 @@ import os
 import os.path
 import pprint
 import sys
+from typing import Optional
 from urllib.parse import urlparse
 
 from mlpstorage_py.benchmarks.base import Benchmark
@@ -811,10 +812,18 @@ class TrainingBenchmark(DLIOBenchmark):
         )
         return int(total_disk_bytes)
 
-    def _capacity_gate_destination(self) -> str:
+    def _capacity_gate_destination(self) -> Optional[str]:
         """Return ``args.data_dir`` — the training dataset destination per
         REQUIREMENTS.md CAP-01.
+
+        Object-mode runs (data_access_protocol == 'object') target an
+        ``s3://`` URI; statvfs against a URI walks to filesystem root and
+        aborts with ``[E401] CAP-01: no valid parent``. Return None to
+        fire the A8 remote-backend escape hatch in ``_pre_execution_gate``.
+        See issue #568.
         """
+        if getattr(self.args, 'data_access_protocol', None) == 'object':
+            return None
         return self.args.data_dir
 
     def datasize(self):
@@ -965,6 +974,11 @@ class CheckpointingBenchmark(DLIOBenchmark):
         """
         cf = getattr(self.args, "checkpoint_folder", None)
         if not cf:
+            return None
+        # Object-mode runs target an s3:// URI; statvfs against a URI
+        # walks to filesystem root and aborts with [E401] CAP-01: no
+        # valid parent. A8 escape hatch — see issue #568.
+        if getattr(self.args, 'data_access_protocol', None) == 'object':
             return None
         return os.path.join(cf, self.args.model)
 
--- a/tests/unit/test_capacity_gate.py
+++ b/tests/unit/test_capacity_gate.py
@@ -40,7 +40,15 @@ import pytest
 # tests/unit/test_benchmarks_kvcache.py.
 import importlib.util as _ilu
 for _dep in ("pyarrow", "pyarrow.ipc", "psutil"):
-    if _ilu.find_spec(_dep) is None and _dep not in sys.modules:
+    if _dep in sys.modules:
+        continue
+    try:
+        _spec = _ilu.find_spec(_dep)
+    except (ModuleNotFoundError, ValueError):
+        # Parent already stubbed above (pyarrow → MagicMock), so find_spec
+        # for the submodule walks into the mock and raises. Treat as missing.
+        _spec = None
+    if _spec is None:
         sys.modules[_dep] = MagicMock()
 
 from mlpstorage_py.benchmarks.base import Benchmark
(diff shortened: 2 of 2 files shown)


russfellows (approved):
Appears correct.

PR #581, opened 2026-06-29
fix(#568): refine CAP-01 object-storage skip — key on storage_type, add helper + dedicated tests

FileSystemGuy:
**Refines** the already-merged #579 fix for #568.

#579 (now in main) shipped a working but narrow fix that keyed on `data_access_protocol == 'object'`. After reviewing PR #574 (closed, no test coverage, broke 2 existing fixtures), this PR combines the best of both:

- Authoritative signal (PR #574's win): keys on `storage.storage_type` so the gate also skips for `--params storage.storage_type=s3` without `--object`.
- Shared helper + DRY (PR #574's structure): `DLIOBenchmark._is_object_storage()` consumed by both `_capacity_gate_destination` overrides.
- Full TDD coverage (#579's discipline): RED commit fails with `AttributeError`, GREEN commit passes; dedicated `TestDLIOIsObjectStorage` class for the helper.

## What changes vs. #579

- `mlpstorage_py/benchmarks/dlio.py` — replaces the two `data_access_protocol`-based checks with a single shared `_is_object_storage()` helper that reads `storage.storage_type` from `params_dict` / `combined_params` (same lookup as `_check_storage_scheme_consistency`) and returns True for `{'s3', 's3_torch'}`. `direct_fs` (`--o-direct`) is correctly NOT skipped — local statvfs still applies.
- `tests/unit/test_capacity_gate.py` — adds `TestDLIOIsObjectStorage` (9 tests covering precedence, fallback, `direct_fs` negative case, `None` combined_params guard, and the no-`--object` edge case). Restructures the two `test_destination_is_none_when_object_storage` tests to use the helper-stub pattern.
- `tests/integration/test_systemname_yaml_end_to_end.py` — stubs `_is_object_storage=False` on the A7-lock fixture (would otherwise silently fail when the helper exists).

## Test plan

RED commit fails with `AttributeError: type object 'DLIOBenchmark' has no attribute '_is_object_storage'` on 9 helper tests + 2 destination tests. GREEN commit passes with no regressions:
- [x] `tests/`: 2392 passed, 13 deselected
- [x] `mlpstorage_py/tests`: 780 passed, 1 xfailed
- [x] `vdb_benchmark/tests`: 144 passed
- [x] `kv_cache_benchmark/tests`: 238 passed
- [ ] Reporter (@jadnov) confirms the original `--params storage.storage_type=s3` repro path also skips the gate cleanly under this refinement.

### DIFF
--- a/mlpstorage_py/benchmarks/dlio.py
+++ b/mlpstorage_py/benchmarks/dlio.py
@@ -381,6 +381,25 @@ class DLIOBenchmark(Benchmark, abc.ABC):
     _OBJECT_URI_SCHEMES = frozenset({'s3', 's3a', 'az', 'gs'})
     _LOCAL_URI_SCHEMES = frozenset({'file', 'direct'})
 
+    def _is_object_storage(self) -> bool:
+        """True when ``storage.storage_type`` resolves to an object backend.
+
+        Reads the same signal as ``_check_storage_scheme_consistency`` —
+        what we actually told DLIO to use after ``_apply_object_storage_params``
+        has run — rather than the user-facing ``data_access_protocol`` CLI
+        positional. That makes the check robust to the
+        ``--params storage.storage_type=s3`` path where a user wires up
+        object storage without passing ``--object``. ``direct_fs`` (the
+        ``--o-direct`` mode) is NOT object storage — it still resolves to
+        a local path and statvfs works.
+        """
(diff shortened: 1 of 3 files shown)


russfellows (approved):
We'll get this working eventually.

PR #577, opened 2026-06-29
fix(#566,#569): CAP-02 shared-FS probe works on FUSE; preserve scp basename

FileSystemGuy:
## Summary

Two tightly coupled fixes that together unblock all multi-host runs on FUSE / distributed filesystems (DAOS, NFS, Lustre, GPFS, BeeGFS, ...). Both surfaced from field reports against `6553c0e`.

- **`mlpstorage_py/cluster_collector.py:2785`** — Drop `st_dev` from the CAP-02 cross-host equality check; compare `st_ino` only. `st_dev` is the kernel's per-mount device id, assigned per-node, so the same shared mount legitimately reports different `st_dev` values on each host. `st_ino` is the cross-host identity signal: if every rank stats rank-0's unique sentinel and sees the same inode, the data-dir is shared. `st_dev` is still surfaced in the failure diagnostic so operators can triage genuinely mismatched mounts.
- **`mlpstorage_py/cluster_collector.py:3076`** — The SSH/SCP staging helper hardcoded the remote destination filename to `mlps_collector.py`. The CAP-02 probe stages a different file (`mlps_cap02_probe.py`), so the script landed at the wrong remote path while `mpirun` pointed at the expected one. Use `basename(script_local_path)` so callers that stage a non-default name reach their script remotely.

Both fixes are required to reach the CAP-02 success path on FUSE in practice — without #569's basename fix the probe never runs remotely; without #566's `st_dev` fix it always fails when it does.

Resolves #566.
Resolves #569.

## Field evidence (from the issues)

```
host=R2-06 rank=0 st_dev=46 st_ino=281482956119445
host=R2-05 rank=1 st_dev=47 st_ino=281482956119445   ← st_dev differs, st_ino IDENTICAL
```

Same pattern reported independently on a 64-node Soperator / Slurm cluster against shared NFS-style mounts.

## Test plan

- [x] Added `TestSharedFsProbeIgnoresStDevAcrossHosts` — exec's the actual `SHARED_FS_PROBE_SCRIPT` with mocked MPI gather returning identical `st_ino` and mismatched `st_dev`; asserts `status="ok"` and `failure_summary=None`. This is the FUSE / DAOS field case from #566.
- [x] Added `TestStageScriptPreservesBasename` — mocks `subprocess.run`; asserts the scp destination ends with the local script basename (`mlps_cap02_probe.py`), not the hardcoded `mlps_collector.py`.
[...]

Issue #569, opened 2026-06-29
Cluster collector remote staging hardcodes the destination filename, breaking the multi-host CAP-02 shared-FS probe


sakib-samar:
# Problem Summary

The collector staging helper `_stage_script_on_remote_hosts` in `mlpstorage_py/cluster_collector.py`, in line 3073–3075, scp's the collector script to each remote host but hardcodes the destination filename:

    scp_cmd = [
        "scp", *ssh_common, script_local_path,
        f"{target}:{remote_dir}/mlps_collector.py",
    ]

This helper has two callers:

- `collect()` stages a script named `mlps_collector.py`, so the hardcode matches and works.
- `run_shared_fs_probe()`, the CAP-02 shared-FS probe, stages `mlps_cap02_probe.py`. The hardcode does not match, so each remote host receives the file as `mlps_collector.py` while `mpirun` invokes `mlps_cap02_probe.py`. Remote ranks can't find the script and the probe fails on every multi-host run.

The destination should mirror the source basename.

# Proposed Solution

```diff
                 scp_cmd = [
                     "scp", *ssh_common, script_local_path,
-                    f"{target}:{remote_dir}/mlps_collector.py",
+                    f"{target}:{remote_dir}/{os.path.basename(script_local_path)}",
                 ]
```

# Testing

Single-host runs unaffected.  On multi-host runs, both the cluster-info collector and the CAP-02 probe scripts land under their correct names on every host, and the probe completes instead of failing on remote ranks.

russfellows:
Can you re-test and reproduce on the CURRENT version?
The most recent Hash-ID is: 6553c0e

sakib-samar:
Thanks @russfellows .

Testing the most recent hash and getting back to you.

sakib-samar:
@russfellows rerpoduced this on https://github.com/mlcommons/storage/commit/6553c0e736b62e0eeedc27cee932419de965fd1e. Added more details on https://github.com/mlcommons/storage/issues/566, as the issues are adjacent.

The runs only succeed after applying patches for both this issue and https://github.com/mlcommons/storage/issues/566.

Issue #567, opened 2026-06-28
--o-direct / storage_type=direct_fs crashes on local NPZ/NPY/JPEG — storage_library sentinel disagreement between config validation ("s3dlio") and the iterable reader ("direct")


jadnov:
Summary
storage_type=direct_fs (reached via mlpstorage's --o-direct) is supposed to route local-filesystem training through the fast iterable data loader using s3dlio's async O_DIRECT engine. In the current DLIO commit it is dead on arrival for the local single-sample formats (NPZ/NPY/JPEG): every run crashes in the warmup batch with FileNotFoundError on a direct:// URI.

The cause is a single magic-string disagreement between two halves of the codebase:

Config validation requires storage_library == "s3dlio" for direct_fs.
The iterable-reader mixin only takes the s3dlio O_DIRECT path when storage_library == "direct".
Because they require different values, there is no value of storage_library that satisfies both — s3dlio passes config but the reader then hands a direct:// URI to plain open(); direct would satisfy the reader but config rejects it.

Environment
dlio_benchmark @ git+https://github.com/mlcommons/DLIO_local_changes.git@252a54b18d113541e1e8e24832921fac0a7f2b96
mlpstorage 3.0.24, s3dlio 0.9.102, torch 2.12.0, Python 3.12, Linux
Workload: unet3d (NPZ), storage_type=direct_fs, local filesystem
Reproduction
mlpstorage closed training unet3d run file --o-direct \
  --num-accelerators 1 --num-client-hosts 1 --hosts <host> \
  --accelerator-type b200 --client-host-memory-in-gb 256 \
  --systemname test \
  --data-dir /mnt/data/unet3d_testset --results-dir /mnt/results \
  --params dataset.num_files_train=10000 --params dataset.num_subfolders_train=5
(any generated UNet3D NPZ dataset on a local/POSIX mount; --o-direct sets storage_type=direct_fs, storage_library=s3dlio, uri_scheme=direct.)

Actual behavior
[ITER_SIMPLE] worker=0 reader=NPZReaderIterable files_this_worker=625 total_files=10000
...
FileNotFoundError: Caught FileNotFoundError in DataLoader worker process 0.
Original Traceback (most recent call last):
  File ".../dlio_benchmark/data_loader/torch_data_loader.py", line 525, in __iter__
    for _batch in self.reader.next():
  File ".../dlio_benchmark/reader/npz_reader_iterable.py", line 47, in next
    yield from self._localfs_stream_next()
[...]

Issue #566, opened 2026-06-27
CAP-02 shared-FS probe rejects all multi-host DAOS/FUSE filesystems due to spurious st_dev comparison


sakib-samar:
# Problem Summary

The CAP-02 shared-FS probe in `mlpstorage_py/cluster_collector.py`, inside the rank-0 analysis block around line 2783, decides whether the data-dir is shared by requiring every rank's (`st_dev`, `st_ino`) tuple to be identical.

This is wrong for FUSE filesystems. `st_dev` is an anonymous device number assigned per mount instance, per node. DAOS DFuse runs a separate daemon on each host, so `st_dev` always differs across nodes even on the same shared mount, making CAP-02 unsatisfiable with no flag or mount-option workaround.

Every rank successfully stats a unique sentinel that only rank 0 created, and `st_ino`, derived from the DAOS object ID, is identical cluster-wide. This can be seen in the run log:

```
host=R2-06 rank=0 st_dev=46 st_ino=281482956119445
host=R2-05 rank=1 st_dev=47 st_ino=281482956119445   ← st_dev differs, st_ino IDENTICAL
```

This blocks all multi-host training and checkpointing runs on any FUSE filesystem, including DAOS, that hit `_pre_execution_gate`.

# Proposed Solution

This can be fixed by dropping `st_dev` from the equality check and comparing `st_ino` only. If all hosts can read the file rank 0 created and report the same inode number, the filesystem is shared. This works correctly on FUSE filesystems like DAOS.

The change is small:

```diff
                 ids = set()
                 for p in all_payloads:
-                    ids.add((p.get("st_dev"), p.get("st_ino")))
+                    ids.add(p.get("st_ino"))
                 if len(ids) != 1:
```

# Testing

Prior to the patch, multi-host runs were resulting in this issue:

```
[root@C-FTC63-R2-06 storage]# ./mlpstorage closed training unet3d datasize \
--systemname my-system \
--hosts C-FTC63-R2-05,C-FTC63-R2-06 \
--num-client-hosts 2 \
--client-host-memory-in-gb 64 \
--max-accelerators 8 --accelerator-type b200 \
--data-dir /mnt/daos/mlps_15c/unet3d/unet3d_data \
--results-dir /mnt/daos/mlps_15c/unet3d/results

Setting attr from max_accelerators to 8
2026-06-27 15:33:00|STATUS:
2026-06-27 15:33:00|STATUS: --- Run Configuration (mlpstorage 3.0.23) ---
[...]

PR #585, opened 2026-06-29
fix(#584): _is_object_storage recognises data_access_protocol=='object'

### LINKED ISSUE #584: `_is_object_storage()` misses the `object` positional (`data_access_protocol`) → object-mode runs fail `[E401] Data directory not found` / checkpoint parent-dir checks
`validation_helpers._is_object_storage(args)` decides whether to skip the local
filesystem path checks, but it only recognizes object mode via **(a)** an explicit
`--params storage.storage_type=s3` or **(b)** an `s3://`-prefixed `data_dir`/`checkpoint_folder`.
It does **not** consult `args.data_access_protocol == 'object'` — the canonical
object-mode signal set by the `object` positional, and the exact signal that the
CAP-01 fix (#568/#579) keys on.

As a result, a perfectly valid object-mode run with a **bare** `--data-dir`
(no `s3://` scheme) and no explicit `--params` is misclassified as local and fails
environment validation with a filesystem error.

## Environment

- mlpstorage `3.0.25` (`origin/main`)
- s3dlio `0.9.102`, DLIO `mlcommons/DLIO_local_changes` rev `252a54b`

## Reproduction

```bash
# .env: BUCKET=<bucket>, AWS_ENDPOINT_URL=…, creds, STORAGE_LIBRARY=s3dlio
mlpstorage closed training unet3d run object \
  --accelerator-type b200 --num-accelerators 8 \
  --client-host-memory-in-gb <N> --data-dir data/unet3d \
  --hosts <host> --systemname <name> --results-dir <init'd-dir> \
  --dlio-bin-path <venv>/bin
```

Result:

```
[E401] Data directory not found: data/unet3d
```

The run is object mode (`object` positional → `data_access_protocol='object'`, `bucket: <bucket>`
is logged), but validation runs `os.path.exists('data/unet3d')` against the local FS.
Today it is only bypassable with `--skip-validation` (which also disables the MPI/SSH/DLIO
checks — too broad).

## Root cause

`mlpstorage_py/validation_helpers.py`:

```python
def _is_object_storage(args) -> bool:
    params = getattr(args, 'params', None) or []
    for p in params:                                   # (a) --params storage.storage_type=s3
        if '=' in p:
            k, v = p.split('=', 1)
            if k.strip() == 'storage.storage_type' and v.strip() in ('s3', 'object'):
                return True
    for attr in ('data_dir', 'checkpoint_folder'):     # (b) s3:// scheme on the path
        val = getattr(args, attr, None)
        if val and str(val).startswith('s3://'):
            return True
    return False                                       # ← misses data_access_protocol == 'object'
```

In `object` mode the `storage.storage_type=s3` param is **injected by `--object` handling
after** argument parsing, so it is not present in `args.params` at validation time; and a
bare `--data-dir`/`--checkpoint-folder` has no `s3://` scheme. So `_is_object_storage`
returns `False`, the early-return guard in `_validate_paths` (~L197) does not fire, and the
local checks at ~L203 (`data_dir`) — and the checkpoint parent-dir check a few lines below —
run against object paths and fail with E401.

## Proposed fix (trivial, `mlcommons/storage`)

Add the canonical signal to `_is_object_storage`, matching what #568 already uses:

```python
    if getattr(args, 'data_access_protocol', None) == 'object':
        return True
```

This makes path validation consistent with the CAP-01 gate (#568) and with the storage
injection, and removes the need for `--skip-validation` on object-mode runs.

I can post a patch but am facing some CLA issues. 

### Verification

`mlpstorage closed training unet3d run object --data-dir data/unet3d …` (bare data-dir,
no `--skip-validation`) should pass environment validation.

## Related

- **#568** [closed] / **#579** [closed] — CAP-01 object-mode skip; **uses `data_access_protocol == 'object'`** (the signal this validator should also use)
- **#581** [open] — *refine CAP-01 object-storage skip — key on storage_type, add helper* (consolidating object-mode detection; the same helper should back path validation)
- `a24c3e0` — commit that introduced `_is_object_storage`
- **#508** [open] — *Enforce --data-dir post-YAML for all training commands* (adjacent `--data-dir` handling)

FileSystemGuy:
## Summary

Resolves #584.

`validation_helpers._is_object_storage(args)` decided object-vs-local using only two signals:

1. `--params storage.storage_type=s3` present in `args.params`
2. `s3://`-scheme prefix on `args.data_dir` / `args.checkpoint_folder`

It did **not** consult `args.data_access_protocol == 'object'` — the canonical signal set by the `object` positional, and the signal every other site already keys on (CAP-01 gate per #568/#579, `dlio.py`, `run_summary`, `cli_parser`, every CLI module).

For a bare `mlpstorage … run object --data-dir data/unet3d …`:

- `data_access_protocol='object'` is set at parse time ✓
- `storage.storage_type=s3` is **injected onto `DLIOBenchmark.params_dict` after** parsing (`mlpstorage_py/benchmarks/dlio.py:202-203`), so it never reaches `args.params` at validation time
- the bare `--data-dir` has no `s3://` scheme

`_is_object_storage` returned `False`, `_validate_paths` fell through to `os.path.exists()` on what is conceptually a bucket key, and the run died with `[E401] Data directory not found`. Only `--skip-validation` could unblock it, which also disables the MPI / SSH / DLIO checks — too broad.

## The fix

One line, additive — check `data_access_protocol` first, then fall through to the legacy signals. Every config that returned `True` before still does.

```python
def _is_object_storage(args) -> bool:
    if getattr(args, 'data_access_protocol', None) == 'object':
        return True
    # … existing legacy signals unchanged …
```

## Tests

12 new tests, all pass; full unit suite stays green.

**`TestIsObjectStorage`** (9 cases):
- canonical signal (`data_access_protocol='object'`) → `True` (the #584 regression lock)
- `data_access_protocol='file'` → `False` (must not misclassify file mode)
- missing `data_access_protocol` attr → falls through to legacy signals
- legacy `--params storage.storage_type=s3` → `True` (and `=object` variant)
- `s3://` on `data_dir` → `True`
- `s3://` on `checkpoint_folder` → `True`
- unrelated `--params` entries don't flip the gate
- empty `Namespace()` doesn't raise

**`TestValidatePathsObjectStorageBypass`** (3 cases — integration through `_validate_paths`):
- exact #584 reproducer config (`object` positional, bare relative `data_dir`, no `--params`) → no errors
- **guardrail:** file mode with a missing `data_dir` still errors (the bypass must not leak)
- checkpoint parent-dir check is also bypassed for object mode (the second site the reporter mentioned)

## Verification

- [x] `uv run pytest tests/unit/test_validation_helpers.py -v` → 46 passed (34 existing + 12 new).
- [x] `uv run pytest tests/unit -q` → **2303 passed**. No regressions.
- [ ] On-cluster: reporter's repro (`mlpstorage closed training unet3d run object --data-dir data/unet3d …`, no `--skip-validation`) should now pass environment validation.

## Notes

- **Independent of #581.** That PR refines a *different* `_is_object_storage` — `DLIOBenchmark._is_object_storage()` in `benchmarks/dlio.py`, scoped to the CAP-01 capacity gate with access to `params_dict`. Same name, different scope. They could converge into a shared helper eventually, but #584 is the user-facing CLI-validation bug and is solved in isolation here.
- Independent of #577 (CAP-02 / scp fixes for #566+#569) and #582 (DLIO #39 pin bump for #567). All three rebase cleanly regardless of merge order.

### DIFF
--- a/mlpstorage_py/validation_helpers.py
+++ b/mlpstorage_py/validation_helpers.py
@@ -165,6 +165,14 @@ def _validate_required_params(args) -> List[Exception]:
 
 def _is_object_storage(args) -> bool:
     """Return True if args indicate object/S3 storage (skip all filesystem checks)."""
+    # Canonical signal: the `object` positional sets data_access_protocol='object'.
+    # storage.storage_type=s3 is injected onto self.params_dict by
+    # DLIOBenchmark._apply_object_storage_params AFTER argument parsing, so it
+    # is not present in args.params at validation time for a bare `--object`
+    # invocation; data_access_protocol IS set at parse time and is what every
+    # other site (CAP-01 gate, dlio.py, run_summary, cli_parser) keys on. See #584.
+    if getattr(args, 'data_access_protocol', None) == 'object':
+        return True
     # Check params list for storage.storage_type=s3
     params = getattr(args, 'params', None) or []
     for p in params:
--- a/tests/unit/test_validation_helpers.py
+++ b/tests/unit/test_validation_helpers.py
@@ -21,6 +21,8 @@ from mlpstorage_py.validation_helpers import (
     _requires_mpi,
     _is_distributed_run,
     _requires_dlio,
+    _is_object_storage,
+    _validate_paths,
 )
 from mlpstorage_py.errors import DependencyError, MPIError, ConfigurationError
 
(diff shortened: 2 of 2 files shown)


russfellows (approved):
Not sure all this is needed to determine if we have object storage, but Ok.

PR #459, opened 2026-06-16
fix(dlio): strip URI scheme from storage.storage_root (#392)

FileSystemGuy:
Resolves #392.

## Summary

When a user passes `--params storage.storage_root=s3://bucket/path`, every S3 upload fails because DLIO's `obj_store_lib` builds URLs with a doubled scheme — `s3://s3://bucket/path/...`. The reporter saw "nothing in S3 / no bucket created" as a result.

The doubled prefix originates in DLIO's `obj_store_lib.ObjStoreLibStorage`:

- `storage_factory.py:49` passes `args.storage_root` straight in as `namespace`.
- `obj_store_lib.py:159` stores it verbatim.
- `obj_store_lib.py:399-401` unconditionally adds `uri_scheme://` when constructing object URIs:
  ```python
  if '://' in str(id):
      return id
  return f"{self.uri_scheme}://{self.namespace.name}/{id.lstrip('/')}"
  ```

So a `storage_root` like `s3://mlperf-data/unet3d` plus a key like `data/unet3d/train/img_000_of_168.npz` produces `s3://s3://mlperf-data/unet3d/data/unet3d/train/img_000_of_168.npz`. MinIO rejects every put, and Keith's command exits without writing anything.

`--object` mode happens to avoid this because `_apply_object_storage_params` sets `storage_root = $BUCKET` (bare bucket, no scheme). Users who hand-roll `--params storage.storage_root=...` hit it.

## What this PR does

- Adds `DLIOBenchmark._strip_uri_scheme` and normalizes `storage.storage_root` in `process_dlio_params` so any `<scheme>://` prefix is stripped before the override is handed off to DLIO. The bare-bucket form is untouched.
- Adds parametrized tests covering `s3://`, `s3://bucket/`, `az://`, `gs://`, and bare-bucket cases, plus the reporter's exact command shape.

The proper fix still belongs in DLIO's `obj_store_lib` — once the pinned fork is patched and `pyproject.toml:127` is bumped, this normalization becomes a harmless no-op, since the normalized input is also valid input.

## Test plan

- [x] `uv run pytest tests/unit -q` → 1372 passed, 4 skipped.
[...]

Issue #392, opened 2026-05-26
Datagen to S3 fails using s3dlio


keithpij:
I am running the datagen command to populate an S3 bucket and the process bombs out. Below is the command I am running. After this command runs there is nothing in S3 (I am using MinIO Free). No bucket is created and there are no objects.  

 mlpstorage training datagen \
  --model unet3d \
  --num-processes 1 \
  --params storage.storage_library=s3dlio \
  --params storage.storage_type=s3 \
  --params storage.storage_root=s3://mlperf-data/unet3d

**The output is below:**

⠋ Validating environment... 0:00:002026-05-26 12:03:10|INFO: Environment validation passed
2026-05-26 12:03:10|STATUS: Benchmark results directory: /tmp/mlperf_storage_results/training/unet3d/datagen/20260526_120310
2026-05-26 12:03:10|WARNING: Results directory not specified. Writing results to the system temp directory: /tmp/mlperf_storage_results. These results will NOT persist across a reboot. Use --results-dir <path> or set the MLPERF_RESULTS_DIR environment variable to save results permanently.
⠋ Validating environment... ━━━━━━━━━━╺━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1/4 0:00:002026-05-26 12:03:10|INFO: MPI BTL transport: auto (OpenMPI default selection)
2026-05-26 12:03:10|STATUS: Running benchmark command:: mpirun -n 1 -host 127.0.0.1:1 --bind-to none --map-by socket /home/keithpij/storage/.venv/bin/dlio_benchmark workload=unet3d_datagen ++hydra.run.dir=/tmp/mlperf_storage_results/training/unet3d/datagen/20260526_120310 ++hydra.output_subdir=dlio_config ++workload.storage.storage_library=s3dlio ++workload.storage.storage_type=s3 ++workload.storage.storage_root=s3://mlperf-data/unet3d --config-dir=/home/keithpij/storage/configs/dlio
[OUTPUT] [DEBUG DLIOBenchmark.__init__] After LoadConfig:
[/home/keithpij/storage/.venv/lib/python3.12/site-packages/dlio_benchmark/utils/utility.py:155]
[OUTPUT]   storage_type   = <StorageType.S3: 's3'> [/home/keithpij/storage/.venv/lib/python3.12/site-packages/dlio_benchmark/utils/utility.py:155]
[...]

PR #586, opened 2026-06-29
fix(#583): bridge URI scheme between DLIO preflight (bare) and checkpoint writer (qualified)

### LINKED ISSUE #583: Object-storage checkpointing double-prefixes the URI (`s3://s3://`) — `checkpoint_folder` never gets the `_strip_uri_scheme` treatment `storage_root` got in #459
Object-mode checkpointing (`mlpstorage closed checkpointing run object …`) cannot
write to an S3 backend. The `checkpoint_folder` URI scheme is mishandled, and **no
form of the argument works**:

- `--checkpoint-folder s3://<bucket>/<prefix>` → the DLIO `ObjStoreLibStorage`
  preflight **prepends the scheme again** → `s3://s3://…` →
  `RuntimeError: Bucket name cannot be empty in URI`.
- `--checkpoint-folder <bucket>/<prefix>` (bare) → the streaming checkpoint
  **writer** rejects it → `Unsupported URI scheme … Supported: file://, direct://, s3://, …`.

This is the **`checkpoint_folder` sibling of #392**: #459 fixed the identical
double-prefix bug for `storage.storage_root` (datagen), but the fix was never
applied to `checkpoint_folder`, and the checkpoint **write** path is also not
scheme-aware.

## Environment

- mlpstorage `3.0.25` (`origin/main`)
- s3dlio `0.9.102`
- DLIO `mlcommons/DLIO_local_changes` rev `252a54b18d113541e1e8e24832921fac0a7f2b96`
- Backend: NVIDIA AIStore S3 endpoint (any S3/MinIO target reproduces it)

## Reproduction

```bash
# .env: BUCKET=<bucket>, AWS_ENDPOINT_URL=…, AWS creds, STORAGE_LIBRARY=s3dlio, S3DLIO_FOLLOW_REDIRECTS=1
mlpstorage whatif checkpointing run object \
  --model llama3-8b --num-processes 8 \
  --checkpoint-folder "s3://$BUCKET/ckpt" \
  --hosts <host> --client-host-memory-in-gb <N> \
  --dlio-bin-path <venv>/bin --systemname <name> \
  --results-dir <init'd-results-dir>
```

Observed (preflight, with the `s3://` form):

```
ConnectionError: ObjStoreLibStorage preflight failed: cannot reach bucket
's3://<bucket>/ckpt/llama3-8b' via s3dlio …
  Underlying error: RuntimeError: Bucket name cannot be empty in URI:
  s3://s3://<bucket>/ckpt/llama3-8b/
```

Observed (writer, with the bare form):

```
[Writer] ERROR: Failed to create storage writer: Unsupported URI scheme:
<bucket>/ckpt/llama3-8b/global_epoch1_step1/zero_pp_rank_0_mp_rank_0_model_states.pt.
Supported: file://, direct://, s3://, az://, gs://
```

## Root cause

Two consumers of `checkpoint_folder` want **opposite** forms, and neither is reconciled:

1. **Preflight wants a *bare* namespace.** DLIO's `ObjStoreLibStorage` preflight builds
   `f"{uri_scheme}://{namespace}/"` (it unconditionally prepends the scheme — the same
   behavior #392 documented). The checkpoint backend's namespace **is** `checkpoint_folder`.
   - `storage.storage_root` is scheme-stripped before becoming a namespace
     (`mlpstorage_py/benchmarks/dlio.py` → `process_dlio_params`, ~L513:
     `normalized = DLIOBenchmark._strip_uri_scheme(storage_root)`, the #459 fix).
   - `checkpoint.checkpoint_folder` is **not** — `add_checkpoint_params`
     (`mlpstorage_py/benchmarks/dlio.py`, ~L903) does
     `os.path.join(self.args.checkpoint_folder, self.args.model)` with no strip.
   - → with `s3://…` it produces `s3://s3://…`.

2. **The writer wants a *scheme-qualified* path.** The checkpoint write path
   (`mlpstorage_py/checkpointing/streaming_checkpoint.py` → `StorageWriterFactory.create(filepath)`,
   ~L370) auto-detects the backend from the URI scheme and rejects a bare path.
   Unlike reads, the write path does **not** go through the storage backend's
   `get_uri()` (which is what prepends the scheme), so a stripped folder leaves the
   writer with no scheme.

So stripping the folder fixes the preflight but breaks the writer; not stripping it
keeps the writer happy but doubles the preflight scheme. Neither form works today.

## Related issues

- **#392** [closed] — *Datagen to S3 fails using s3dlio* (the original double-prefix report, for `storage_root`)
- **#459** [closed] — *fix(dlio): strip URI scheme from `storage.storage_root` (#392)* (the fix this issue extends to `checkpoint_folder`)
- **#538** [closed] / **#553** [closed] — fail-fast on `storage_type` vs folder **scheme mismatch** (local-vs-object; different from this in-object double-prefix)
- **#568** [closed] / **#574** [closed] / **#581** — CAP-01 `statvfs` on `s3://` (a separate object-mode gate bug, already fixed)

FileSystemGuy:
Fixes #583. Recreates @gaikwadabhishek's analysis as a CLA-clean implementation — credit for the bug report and root-cause writeup is his.

## The bug

`mlpstorage closed checkpointing run object --checkpoint-folder s3://…` cannot write to S3. Two consumers of `checkpoint_folder` want opposite forms:

- **DLIO preflight** (`ObjStoreLibStorage._preflight`) prepends the scheme itself (the #392 pattern), so it wants a **bare** bucket/prefix. Feeding it `s3://…` produces `s3://s3://…` and aborts.
- **mlpstorage's streaming-checkpoint writer** (`S3DLIOStorageWriter`) auto-dispatches by URI scheme, so it wants the **scheme-qualified** form. A bare path either lands silently on `FileStorageWriter` (writing to the local FS instead of S3 — data integrity failure) or raises `Unsupported URI scheme`.

Neither workaround works today — both shapes break one of the two consumers.

## The fix

Mirror the #459 storage_root fix for `checkpoint_folder`, plus a one-env-var bridge to the writer subprocess:

1. **`CheckpointingBenchmark.add_checkpoint_params`** — strip the URI scheme from `checkpoint_folder` so DLIO preflight gets a bare namespace. Set `MLPSTORAGE_CHECKPOINT_URI_SCHEME=<scheme>` in the parent env. Conditioned on `storage.storage_type in {'s3','s3_torch'}` so the file-mode mismatch guardrail in `_check_storage_scheme_consistency` (which catches `--file --checkpoint-folder s3://…` user mistakes) keeps working. `direct_fs` (`--o-direct`) is unaffected — its namespace is a local path.
2. **`storage_writers/_normalize_checkpoint_uri`** — new helper. If URI lacks a scheme AND env var is set, prepend `{scheme}://`. No-op otherwise.
3. **`StorageWriterFactory.create`** and **`S3DLIOStorageWriter.__init__`** — both call the helper before dispatch so the writer subprocess (forked from DLIO, inherits env) reconstructs the qualified URI cleanly.

## Test plan

- [x] RED commit (5d2b5cc) adds 21 tests across `tests/unit/test_dlio_object_storage.py` (10 new in `TestAddCheckpointParamsSchemeStripping`) and a new `tests/unit/test_checkpoint_writer_scheme.py` (11 tests: 6 for the helper, 4 for factory, 2 for writer init). 14 fail with `AttributeError` / `ImportError` / wrong dispatch; 7 pass (negative-case regression guards).
- [x] GREEN commit (22a99d4) makes all 21 pass. No regressions:
  - `tests/`: 2406 passed, 13 deselected
  - `mlpstorage_py/tests`: 780 passed, 1 xfailed
  - `vdb_benchmark/tests`: 144 passed
  - `kv_cache_benchmark/tests`: 238 passed
- [ ] On-cluster: reporter's repro (`mlpstorage closed checkpointing run object --checkpoint-folder s3://… --model llama3-8b …`) writes a full checkpoint to the bucket without `[E401]` or `Unsupported URI scheme`.

## Relationship to other open PRs

End-to-end object-mode checkpointing also requires:
- **#585** (`validation_helpers._is_object_storage` fix for #584) — without it the `[E401] Data directory not found` validation error fires before reaching this code path. Independent fix, different file.
- **#581** (CAP-01 object-storage skip refinement for #568) — uses the same `storage.storage_type` signal as this PR but inside `DLIOBenchmark._is_object_storage()`. This PR's inline check could refactor to call that helper once both land. Different functions in the same file; rebases cleanly either order.

All three rebase cleanly regardless of merge order; together they unblock object-mode runs end-to-end.

## Files

- `mlpstorage_py/benchmarks/dlio.py` — `add_checkpoint_params` scheme strip + env var; import for `CHECKPOINT_URI_SCHEME_ENV`.
- `mlpstorage_py/checkpointing/storage_writers/__init__.py` — new `_normalize_checkpoint_uri` helper + factory wiring; `CHECKPOINT_URI_SCHEME_ENV` constant.
- `mlpstorage_py/checkpointing/storage_writers/s3dlio_writer.py` — call the helper in `__init__`.
- `tests/unit/test_dlio_object_storage.py` — new `TestAddCheckpointParamsSchemeStripping` (10 tests).
- `tests/unit/test_checkpoint_writer_scheme.py` — new file (3 test classes, 11 tests).

### DIFF
--- a/mlpstorage_py/benchmarks/dlio.py
+++ b/mlpstorage_py/benchmarks/dlio.py
@@ -7,6 +7,7 @@ from typing import Optional
 from urllib.parse import urlparse
 
 from mlpstorage_py.benchmarks.base import Benchmark
+from mlpstorage_py.checkpointing.storage_writers import CHECKPOINT_URI_SCHEME_ENV
 from mlpstorage_py.config import (CONFIGS_ROOT_DIR, BENCHMARK_TYPES, EXEC_TYPE, MPIRUN, MLPSTORAGE_BIN_NAME,
                                LLM_ALLOWED_VALUES, LLM_SUBSET_PROCS, EXIT_CODE, MODELS, HYDRA_OUTPUT_SUBDIR,
                                LLM_SIZE_BY_RANK)
@@ -900,7 +901,26 @@ class CheckpointingBenchmark(DLIOBenchmark):
             # For direct:// / file:// schemes, ObjStoreLibStorage._preflight validates
             # that namespace as a local directory — so it must be an absolute path,
             # not relative to storage.storage_root. See issue #536.
-            self.params_dict['checkpoint.checkpoint_folder'] = os.path.join(self.args.checkpoint_folder, self.args.model)
+            folder = self.args.checkpoint_folder
+            storage_type = self.params_dict.get('storage.storage_type', 'local')
+            if storage_type in DLIOBenchmark._OBJECT_STORAGE_TYPES:
+                # Issue #583: mirror the #459 storage_root fix — DLIO's
+                # ObjStoreLibStorage._preflight prepends the scheme itself,
+                # so feeding it s3://bucket/... produces s3://s3://bucket/...
+                # Strip the scheme and stash it in an env var so the
+                # streaming-checkpoint writer (forked from DLIO, inherits
+                # env) can reconstruct the qualified URI at dispatch time.
+                normalized = DLIOBenchmark._strip_uri_scheme(folder)
+                if normalized != folder:
+                    scheme = urlparse(folder).scheme
+                    os.environ[CHECKPOINT_URI_SCHEME_ENV] = scheme
+                    self.logger.debug(
+                        f'Normalized checkpoint_folder: {folder!r} -> {normalized!r}; '
+                        f'set {CHECKPOINT_URI_SCHEME_ENV}={scheme!r} for the '
+                        'streaming-writer subprocess (issue #583).'
+                    )
+                folder = normalized
+            self.params_dict['checkpoint.checkpoint_folder'] = os.path.join(folder, self.args.model)
 
 
     def add_workflow_to_cmd(self, cmd) -> str:
--- a/mlpstorage_py/checkpointing/storage_writers/__init__.py
+++ b/mlpstorage_py/checkpointing/storage_writers/__init__.py
@@ -12,6 +12,8 @@ Use StorageWriterFactory.create() to automatically select the appropriate
 backend based on URI scheme or explicit backend name.
 """
 
+import os
+
 from .base import StorageWriter
 from .file_writer import FileStorageWriter
 from .s3dlio_writer import S3DLIOStorageWriter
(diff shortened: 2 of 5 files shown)


russfellows (approved):
Last one for today... maybe.

PR #587, opened 2026-06-29
fix(#573): CAP-02 probe wire payload exceeds PIPE_BUF at scale — drop ranks array + cap failure messages

### LINKED ISSUE #573: CAP-02: rank-0 result exceeds PIPE_BUF at scale → truncated stdout → `payload unreadable`
## Summary
After fixing bug 1, the probe runs but at larger node counts the launcher fails:

```
[E404] CAP-02: shared-FS probe rank-0 payload unreadable:
       Expecting ',' delimiter: line 1 column 4096 (char 4095)
```

The probe (Step J) emits its result as a **single-line JSON** on stdout between
`__CAP02_RESULT_BEGIN__`/`__CAP02_RESULT_END__`, and that JSON embeds the full
per-rank gather:

```python
_result = {"status": ..., "ranks": all_payloads, "failure_summary": ..., "unlink_warning": ...}
print(json.dumps(_result, separators=(",", ":")), flush=True)   # "stay under PIPE_BUF"
```

`all_payloads` is one dict per rank (`hostname, rank, failure, st_dev, st_ino`).
At ~64 ranks this line exceeds **PIPE_BUF (4096 bytes)**, so the write is no
longer atomic and the I/O-forwarding layer truncates it at 4096 — the launcher's
marker regex then captures a truncated line and `json.loads` fails at char 4095.
Works at ≤~32 nodes (JSON < 4096), fails at 64.

## Fix
The launcher only consumes `status` and `failure_summary` (it never reads
`ranks`). Drop the per-rank array from the emitted result so the line stays
small at any scale:

```python
_result = {"status": ..., "failure_summary": rank0_failure_summary, "unlink_warning": ...}
```

(`failure_summary` is already a compact human string built from `all_payloads`
in Step E; on the success path it is `None`. If the failure message itself could
exceed PIPE_BUF on huge clusters, cap it too — e.g. list only the first K
failing ranks.)

## Environment / Impact
Same as bug 1 (mlpstorage 3.0.24, Soperator, 64 nodes). Blocks all multi-node
runs at ~≥40-node scale regardless of filesystem. Worked around locally by
dropping `ranks` from the probe's stdout result.

FileSystemGuy:
Fixes #573. Credit to @jadnov for the bug report and root-cause writeup.

## The bug

At >=~40 ranks the CAP-02 shared-FS probe aborts every multi-node run with:

```
[E404] CAP-02: shared-FS probe rank-0 payload unreadable:
       Expecting ',' delimiter: line 1 column 4096 (char 4095)
```

Rank 0 emits its result as a single compact JSON line between `__CAP02_RESULT_BEGIN__` / `__CAP02_RESULT_END__` markers, and the payload embeds the full per-rank gather under a `"ranks"` key (one ~60-byte dict per rank). At 64 ranks the line crosses **PIPE_BUF (4096 B)**, the I/O forwarding layer stops writing atomically, the launcher's marker regex captures a truncated line, and `json.loads` fails at exactly column 4095.

Works at ≤~32 ranks (JSON < 4096), fails at 64+. Same root cause as the underlying kernel pipe semantics, not the application code.

### 1. Drop `"ranks"` from the wire entirely

The launcher reads only `status`, `failure_summary`, `unlink_warning` (cluster_collector.py:3625-3631). The per-rank array had zero consumers — pure wire overhead. Per-rank failure detail is already flattened into `failure_summary.message` by `_build_per_rank_message` and `_build_cardinality_message` before serialization, so operator-visible content is preserved verbatim (host/rank/mode/errno/message for failed ranks; host/rank/st_dev/st_ino for cardinality faults).

### 2. Cap the failure-message builders at 16 per-host lines

The reporter explicitly flagged this follow-up: even with `"ranks"` dropped, `failure_summary.message` at a 256-rank cardinality mismatch is ~15 KB — re-introducing the same truncation through a different builder. Capping at 16 lines + `... and N more ranks omitted` keeps the wire bounded at any rank count while preserving operator visibility into the scale of the failure. Small clusters (≤16 ranks) see no change.

### Side effects

- **Integration test updated.** `test_two_local_ranks_same_tmpfs_succeeds_silently` asserted on `parsed['ranks']` to re-verify cardinality. Switched to `status == 'ok'` + `failure_summary is None` — the launcher contract. The cardinality check still runs on rank 0 before serialization; status alone is sufficient proof of success.
- **Drive-by test hygiene.** `TestVectorDBBenchmark.test_constructor_accepts_kwargs` and `test_datasize_does_not_require_pymilvus` used hardcoded `results_dir='/tmp/test'` + fixed `run_datetime`. After ~10 dev runs the reserve-directory retry window (10 slots) is exhausted and the tests fail. Switched to per-test `tmp_path`.

## Test plan

Two RED commits (c1c97df, b642f74), one GREEN commit (b964c0f). 10 new tests across two classes in `tests/unit/test_cluster_collector.py`:

**`TestRank0JsonStaysUnderPipeBuf`** (3 tests via probe-script exec):
- `test_emitted_json_excludes_ranks_array` — shape lock on the wire schema.
- `test_wire_line_stays_under_pipe_buf_at_256_ranks` — primary size lock; current bug.
- `test_failure_path_also_excludes_ranks_array` — shape lock on the failure leg.

**`TestFailureMessagesCappedAtScale`** (4 tests):
- `test_cardinality_message_lists_every_host_on_small_cluster` — cap is a safety net, not unconditional trim.
- `test_per_rank_failure_at_256_ranks_stays_under_pipe_buf` — defensive cap on the per-rank builder.
- `test_cardinality_failure_at_256_ranks_stays_under_pipe_buf` — end-to-end size lock on the cardinality leg.
- `test_cardinality_failure_at_256_ranks_announces_elision` — operator-visibility lock.

Plus 1 updated integration test (`test_two_local_ranks_same_tmpfs_succeeds_silently`) and 2 updated VDB-test fixtures (drive-by hygiene).

Verification:
- [x] `tests/`: 2414 passed, 13 deselected
- [x] `mlpstorage_py/tests`: 780 passed, 1 xfailed
- [x] `vdb_benchmark/tests`: 144 passed
- [x] `kv_cache_benchmark/tests`: 238 passed
- [ ] On-cluster: reporter's repro at 64 nodes — CAP-02 probe should complete without `[E404] payload unreadable`.

## Files

- `mlpstorage_py/cluster_collector.py` — drop `"ranks"` from `_result`; add 16-line cap + omitted-count tail in both `_build_cardinality_message` and `_build_per_rank_message`; update schema docstring.
- `tests/unit/test_cluster_collector.py` — two new test classes (10 tests).
- `tests/integration/test_shared_fs_probe_real_mpi.py` — update success-case assertion to use the launcher contract.
- `mlpstorage_py/tests/test_benchmarks.py` — per-test `tmp_path` for two VDB constructor tests.

## Overlap with main

PR #577 (FUSE `st_dev`/`st_ino` fix for #566+#569) just merged and touches the same function (line ~2784, the cardinality check). My changes are at lines ~2611 (message builders) and ~2845 (`_result` dict) — different hunks. Rebased cleanly on current `main`.

### DIFF
--- a/mlpstorage_py/cluster_collector.py
+++ b/mlpstorage_py/cluster_collector.py
@@ -2572,16 +2572,19 @@ if __name__ == '__main__':
 # JSON output schema (rank 0 only, between BEGIN/END markers):
 #   {
 #     "status": "ok" | "fail",
-#     "ranks": [
-#       {"hostname": str, "rank": int, "failure": None | dict, "st_dev": int|None, "st_ino": int|None},
-#       ...
-#     ],
 #     "failure_summary": None | {
 #       "kind": "cardinality" | "per_rank",
-#       "message": str   # human-readable, used by the launcher verbatim
+#       "message": str   # human-readable, used by the launcher verbatim;
+#                        # built from per-rank payloads in Step E so the
+#                        # detail survives without shipping the array.
 #     },
 #     "unlink_warning": None | str    # set if rank-0 unlink failed (D-44 cosmetic)
 #   }
+#
+# Issue #573: the per-rank gather (``all_payloads``) is NOT emitted on
+# the wire. The launcher never consumed it, and at >=~40 ranks the
+# array pushed the JSON line past PIPE_BUF (4096 B), causing
+# non-atomic stdout truncation and a json.loads failure at char 4095.
 #
 # Exit codes:
 #   0 on status='ok'
@@ -2605,6 +2608,10 @@ import sys
 import time
 
 
+_PER_HOST_LINE_CAP = 16  # Issue #573: sample-plus-summary cap so the
+                         # wire JSON stays under PIPE_BUF at any scale.
+
+
 def _build_cardinality_message(payloads):
     """Build the verbatim multi-line error body for a cardinality > 1 fault.
 
@@ -2617,13 +2624,19 @@ def _build_cardinality_message(payloads):
         ...
       This typically means one or more hosts have a local-disk path where a
       shared mount was expected.
+
+    Issue #573: per-host lines are capped at _PER_HOST_LINE_CAP. Above
+    the cap, a "... and N more ranks omitted" tail line tells the
+    operator the true scale of the failure without pushing the wire
+    JSON over PIPE_BUF.
     """
     lines = []
     lines.append(
         "CAP-02: shared-FS probe detected the data-dir is NOT the same "
         "filesystem on every participating host."
     )
-    for p in payloads:
+    shown = payloads[:_PER_HOST_LINE_CAP]
+    for p in shown:
         lines.append(
             "  host={h} rank={r} st_dev={d} st_ino={i}".format(
                 h=p.get("hostname", "?"),
@@ -2632,6 +2645,9 @@ def _build_cardinality_message(payloads):
                 i=p.get("st_ino"),
             )
         )
+    omitted = len(payloads) - len(shown)
+    if omitted > 0:
+        lines.append("  ... and {n} more ranks omitted (issue #573 PIPE_BUF cap)".format(n=omitted))
     # REQUIREMENTS.md CAP-02 verbatim hint lock (single-line literal so the
     # plan's `grep -c '...'` acceptance criterion matches exactly):
     lines.append("this typically means one or more hosts have a local-disk path where a shared mount was expected.")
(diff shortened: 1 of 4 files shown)


russfellows (approved):
Looks thoroughly tested.

PR #602, opened 2026-06-30
KV cache Rules for closed and open submission

hazemawadalla:


dslik:
@FileSystemGuy There seems to be some sort of automated rules checking going on. Can you update the system that drives this to be compatible with @hazemawadalla's additions?

FileSystemGuy:
@dslik I don't understand, what does "to be compatible with" mean?  What action are you thinking should happen but isn't, or what action is failing that you think should happen?

dslik:
> @dslik I don't understand, what does "to be compatible with" mean? What action are you thinking should happen but isn't, or what action is failing that you think should happen?

The automated tests check every section of the rules against existing tests (or want OUT_OF_SCOPE_RULES marked), causing the tests to fail against this check-in, even though it only modifies a markdown file.

FileSystemGuy:
@dslik Ah, yes the new mlpstorage checks the Rules.md file for new requirements that it has not been told about.  That's supposed to force me to not forget to build out the mlpstorage support for any rules changes.  I'll dig into it in more detail after the evening WG meeting.

FileSystemGuy:
@hazemawadalla @dslik Claude has noticed that:

1. Internal contradiction on inter-option-delay. §6.3.2.1 declares --inter-option-delay = 90 s, but §6.6.1 references "sequence locks of 6.3.2.1 (seed 42, trials 3, inter-option-delay 20 s, no --config)." Only one can be right.

2. Rules disagree with code. mlpstorage_py/benchmarks/kvcache.py:247 hard-fails any CLOSED run whose inter_option_delay != 20, and the CLI default is 20 (cli/kvcache_args.py:214). §6.6.1's 20 s matches the shipped code; §6.3.2.1's 90 s does not. If 90 s is the actual intent, kvcache.py:247–262 must change too.

Could you please confirm that --inter-option-delay = 20 is the correct value for CLOSED and the correct default for OPEN?

FileSystemGuy:
Two spec clarifications from offline discussion, so they're recorded here for reviewers:

**1. `--inter-option-delay = 20 seconds` (CLOSED).**
§6.3.2.1 in this PR currently reads `--inter-option-delay = 90 s`; §6.6.1 correctly says 20 s. The shipped code (`mlpstorage_py/benchmarks/kvcache.py:247` hard-fails on any CLOSED value ≠ 20; `mlpstorage_py/cli/kvcache_args.py:214` defaults to 20) is authoritative. Please update §6.3.2.1 to 20 s so it matches §6.6.1 and the code.

**2. Four-part §6.3 IDs (`6.3.1.1`, `6.3.2.1`, `6.3.4.5`, etc.) are intentional and stay.**
The current `rules_coverage.py:54` discovery regex only matches 3-part IDs, so it silently skips the twelve §6.3 rules — which is why this PR's CI failure lists only 7 unmapped IDs (§6.4/§6.5/§6.6) instead of 19. The §6.3 rules aren't unmapped, they're undiscovered; without a regex fix, the coverage gate can never enforce them.

I filed #653 to relax the regex to accept N-part IDs. That should be merged first, then this PR's CI will surface the full 19-rule unmapped set, and we can decide per-rule between real `@rule` bindings and `OUT_OF_SCOPE_RULES` entries (my guess: 6.4.2 and 6.6.3 are OPEN-narrative and belong out-of-scope with a reason; the rest should get real bindings — but that's a follow-on PR, not something this doc-only PR needs to carry).

TL;DR to unblock merge:
- Change §6.3.2.1 from 90 s → 20 s.
- Wait for #653 to land (one-line regex change).
- Then a small companion PR on `main` populates `coverage_mapping.py` for the 19 IDs so the CI test passes.

Happy to draft either the #653 fix or the coverage_mapping companion once the WG signs off on the direction.

hazemawadalla:
Hey Curtis,
I'm confused why do we need a test for changes to a markdown file ? I mean it seems like a good feature it's just odd . I thought we said the inter run delay is 90s

FileSystemGuy:
@hazemawadalla @dslik If you want 90 instead of 20, I'll change it, I just needed to know which one you want.

The new mlpstorage command cross checks all the rules in Rules.md against the code that implements those rules in the validation checker.  Its good s/w engineering.

PR #613, opened 2026-06-30
Fix report generation to avoid all results grouped in same JSON

wolfgang-desalvador:
This pull request refactors the report generation logic so that summary files (`results.json` and `results.csv`) are now written inside each discovered `run/` folder, rather than in a global output directory. The tests and file-writing methods have been updated to reflect this new per-run-folder rollup behavior.

**Report generation changes:**

* The `ReportGenerator` now writes `results.json` and `results.csv` files into each `run/` folder containing benchmark run results, rather than honoring the `--output-dir` flag or writing to a global directory. This ensures that rollups are colocated with their respective runs and only include the timestamps present in each folder. (`mlpstorage_py/report_generator.py`) [[1]](diffhunk://#diff-309065fdcd438918fd57a5708268b73fd4c00d6527e968fe8176680d04455381L86-R89) [[2]](diffhunk://#diff-309065fdcd438918fd57a5708268b73fd4c00d6527e968fe8176680d04455381L200-R229)
* The `write_json_file` and `write_csv_file` methods now accept an optional `target_dir` argument to specify the output directory, defaulting to the per-run-folder location. (`mlpstorage_py/report_generator.py`)

**Test updates:**

* Unit tests have been updated to create mock run directories that match the expected on-disk layout, and assertions now check for rollup files inside the appropriate `run/` folders rather than at the root. (`tests/unit/test_reporting.py`) [[1]](diffhunk://#diff-ad70d771bd9b2ade956b089d909d711788cba84d4a4690d019ad4006b3ade965R215-R228) [[2]](diffhunk://#diff-ad70d771bd9b2ade956b089d909d711788cba84d4a4690d019ad4006b3ade965L248-R269) [[3]](diffhunk://#diff-ad70d771bd9b2ade956b089d909d711788cba84d4a4690d019ad4006b3ade965R508-R521) [[4]](diffhunk://#diff-ad70d771bd9b2ade956b089d909d711788cba84d4a4690d019ad4006b3ade965L528-R549)

### DIFF
--- a/mlpstorage_py/report_generator.py
+++ b/mlpstorage_py/report_generator.py
@@ -83,13 +83,10 @@ class ReportGenerator:
         # No-op when --results-dir already points at a flat benchmark-type root.
         self.results_dir = self._resolve_effective_results_dir(self.results_dir)
 
-        # Honor --output-dir for write_{json,csv}_file (falls back to results_dir).
-        output_dir = None
-        if self.args is not None:
-            output_dir = getattr(self.args, 'output_dir', None)
-        self.output_dir = output_dir or self.results_dir
-        if self.output_dir and self.output_dir != self.results_dir:
-            os.makedirs(self.output_dir, exist_ok=True)
+        # NOTE: results.json / results.csv are written into each discovered
+        # <...>/run/ folder, one rollup per run/ folder containing only the
+        # timestamps inside that folder. --output-dir is intentionally not
+        # honored — the rollup location is dictated by the on-disk layout.
 
         # Initialize formatters
         self.msg_formatter = ValidationMessageFormatter(use_colors=use_colors)
@@ -197,15 +194,39 @@ class ReportGenerator:
         # Verify the results directory exists:
         self.logger.info(f'Generating reports for {self.results_dir}')
 
-        # Always traverse the full directory and emit one rollup
-        # results.json / results.csv covering every discovered run, written
-        # to self.output_dir (which defaults to self.results_dir when
-        # --output-dir is not supplied).
-        run_result_dicts = [
-            report.benchmark_run.as_dict() for report in self.run_results.values()
-        ]
-        self.write_csv_file(run_result_dicts)
-        self.write_json_file(run_result_dicts)
+        # Group accumulated runs by their parent `run/` folder. Each leaf
+        # `benchmark_run.result_dir` is `<...>/<benchmark>/<model>/run/<timestamp>/`;
+        # the rollup is written INSIDE the parent `<...>/<benchmark>/<model>/run/`
+        # directory and contains only the timestamps that live in that
+        # specific `run/` folder. One results.json + one results.csv per
+        # run/ folder; nothing is written elsewhere in the tree.
+        groups: Dict[str, List[dict]] = {}
+        skipped = 0
+        for result in self.run_results.values():
+            leaf = getattr(result.benchmark_run, 'result_dir', None)
+            if not leaf:
+                self.logger.warning(
+                    "Run %s has no result_dir on its BenchmarkRun; "
+                    "cannot place a rollup next to it. Skipping.",
+                    result.benchmark_run.run_id,
+                )
+                skipped += 1
+                continue
+            run_folder = os.path.dirname(os.path.abspath(leaf))
+            groups.setdefault(run_folder, []).append(
+                result.benchmark_run.as_dict()
+            )
+
+        if not groups:
+            self.logger.warning(
+                "No run/ folders to write rollups for (skipped %d runs without result_dir).",
+                skipped,
+            )
+            return EXIT_CODE.SUCCESS
+
+        for run_folder, run_dicts in sorted(groups.items()):
+            self.write_json_file(run_dicts, target_dir=run_folder)
+            self.write_csv_file(run_dicts, target_dir=run_folder)
 
         return EXIT_CODE.SUCCESS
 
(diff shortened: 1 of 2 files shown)


wolfgang-desalvador:
@FileSystemGuy  please check this, since fixes an issues introduced with the last merge. Results are collecting all benchmarks in each run folder, with a lot of duplication. 

This should restore the functionality

FileSystemGuy:
@wolfgang-desalvador Thanks for looking so closely at the benchmark and noticing things that don't make sense!  This is what we really need a lot more of.

This codebase started life with several (human) authors, so there isn't really very much internal documentation on what is happening and why.  In this case, there are reasons the code does what it does, they just not obvious reasons.  :-( 

The Training workloads run the actual workload 6 times, where the first run is a throwaway warmup run, only the 5 following runs are part of the submission results.  That's a deliberate design choice (which certainly needs documenting in the code so that future spelunkers like yourself don't get led astray!).  A single checkpointing run is composed of writing 10 large files, then reading those 10 files back, and since the first step is writing, there's no need for or value in a warmup run, so we don't do one for checkpointing.

The core of the confusion is that some piece of code (likely in DLIO rather than mlpstorage) is deliberately changing the timestamp on the first run of the 6 to be the timestamp from the second run of the 6.  Note that all other per-run timestamps match the directory name except the first run.

Here's an example `reportgen` from a v2.0 submission cross-referenced disk pathname vs. JSON:

| Workload | `run/<ts>/` dirs on disk | Records in `results.json` |
|---|---:|---:|
| training/cosmoflow | 6 | 5 |
| training/unet3d | 6 | 5 |
| training/resnet50 | 6 | 5 |
| checkpointing/llama3-8b | 1 | 1 |
| checkpointing/llama3-70b | 1 | 1 |
| checkpointing/llama3-405b | 1 | 1 |
| checkpointing/llama3-1t | 2 | 2 |

We need to change the output from the result-tree-walker to eliminate the confusion that there's duplication and/or **data loss** (Claude told me that in a horrified tone! :-), so here's what I suggest:

```text
Training - resnet50, Accelerator: h100
      Runs:
        - training_run_resnet50_2025-07-10T19:56:24 [WARMUP, not aggregated]
        - training_run_resnet50_2025-07-10T20:06:28 [CLOSED]
        - training_run_resnet50_2025-07-10T20:29:12 [CLOSED]
        ...
```

The reportgen functionality is ultimately intended to summarize everything that is in the --results-dir so that the submitter and the MLCommons staff member who receives the benchmark results upload file can see everything that they're submitting as benchmark results in one place.  Ie: "everything is squashed together" is a feature, not a bug.

In that context, if the user runs reportgen, then does some more benchmarking, then runs reportgen again, it should just overwrite the two output files to reflect the current contents of the --results-dir.1

WRT the --output-directory flag, I'm actually inclined to simplify life by de-supporting that flag, that the only place the "mlpstorage reports reportgen --results-dir /tmp/nerf" output files can go is at the root of the submission package as a whole (eg: /tmp/nerf/results.{csv,json}).

Does this all make sense?  Is your "BS meter" going off because I've missed something?  Is there a better plan?

I propose that we do not merge this PR, that I submit a PR that changes the walker output files as shown above, and eliminate the --output-directory flag in favor of a fixed location.  Does that sound right?

Again, thank you for looking at the benchmark and trying to make it better!

wolfgang-desalvador:
> @wolfgang-desalvador Thanks for looking so closely at the benchmark and noticing things that don't make sense! This is what we really need a lot more of.
> 
> This codebase started life with several (human) authors, so there isn't really very much internal documentation on what is happening and why. In this case, there are reasons the code does what it does, they just not obvious reasons. :-(
> 
> The Training workloads run the actual workload 6 times, where the first run is a throwaway warmup run, only the 5 following runs are part of the submission results. That's a deliberate design choice (which certainly needs documenting in the code so that future spelunkers like yourself don't get led astray!). A single checkpointing run is composed of writing 10 large files, then reading those 10 files back, and since the first step is writing, there's no need for or value in a warmup run, so we don't do one for checkpointing.
> 
> The core of the confusion is that some piece of code (likely in DLIO rather than mlpstorage) is deliberately changing the timestamp on the first run of the 6 to be the timestamp from the second run of the 6. Note that all other per-run timestamps match the directory name except the first run.
> 
> Here's an example `reportgen` from a v2.0 submission cross-referenced disk pathname vs. JSON:
> 
> Workload	`run/<ts>/` dirs on disk	Records in `results.json`
> training/cosmoflow	6	5
> training/unet3d	6	5
> training/resnet50	6	5
> checkpointing/llama3-8b	1	1
> checkpointing/llama3-70b	1	1
> checkpointing/llama3-405b	1	1
> checkpointing/llama3-1t	2	2
> We need to change the output from the result-tree-walker to eliminate the confusion that there's duplication and/or **data loss** (Claude told me that in a horrified tone! :-), so here's what I suggest:
> 
> ```
> Training - resnet50, Accelerator: h100
>       Runs:
>         - training_run_resnet50_2025-07-10T19:56:24 [WARMUP, not aggregated]
>         - training_run_resnet50_2025-07-10T20:06:28 [CLOSED]
>         - training_run_resnet50_2025-07-10T20:29:12 [CLOSED]
>         ...
> ```
> 
> The reportgen functionality is ultimately intended to summarize everything that is in the --results-dir so that the submitter and the MLCommons staff member who receives the benchmark results upload file can see everything that they're submitting as benchmark results in one place. Ie: "everything is squashed together" is a feature, not a bug.
> 
> In that context, if the user runs reportgen, then does some more benchmarking, then runs reportgen again, it should just overwrite the two output files to reflect the current contents of the --results-dir.1
> 
> WRT the --output-directory flag, I'm actually inclined to simplify life by de-supporting that flag, that the only place the "mlpstorage reports reportgen --results-dir /tmp/nerf" output files can go is at the root of the submission package as a whole (eg: /tmp/nerf/results.{csv,json}).
> 
> Does this all make sense? Is your "BS meter" going off because I've missed something? Is there a better plan?
> 
> I propose that we do not merge this PR, that I submit a PR that changes the walker output files as shown above, and eliminate the --output-directory flag in favor of a fixed location. Does that sound right?
> 
> Again, thank you for looking at the benchmark and trying to make it better!

Thank you @FileSystemGuy. The other PR looks great. I have created another PR [here](https://github.com/mlcommons/storage/pull/620/changes) addressing one lost open point. Commented on the other PR also

wolfgang-desalvador:
Not needed since duplicated

PR #606, opened 2026-06-30
Patch submission checker for v3.0

wolfgang-desalvador:
This pull request introduces several important improvements and bug fixes to the report generation and submission checking logic for mlpstorage. The most significant changes are the introduction of robust directory resolution in the report generator, improved handling of output directories, updates to parameter extraction to match new metadata conventions, and adjustments to error/warning reporting to avoid blocking submissions on known upstream issues.

**Report Generation Improvements**
- Enhanced the `ReportGenerator` to automatically detect and resolve the correct results directory, supporting both legacy and canonical mlpstorage submission tree layouts. This ensures reports are generated for the intended system and benchmark-type directories, improving compatibility and correctness. (`mlpstorage_py/report_generator.py`) [[1]](diffhunk://#diff-309065fdcd438918fd57a5708268b73fd4c00d6527e968fe8176680d04455381R79-R92) [[2]](diffhunk://#diff-309065fdcd438918fd57a5708268b73fd4c00d6527e968fe8176680d04455381R110-R166)
- Added support for specifying an `--output-dir` for generated report files, defaulting to the resolved results directory if not provided. Output files (`results.json`, `results.csv`) are now consistently written to this location. (`mlpstorage_py/report_generator.py`) [[1]](diffhunk://#diff-309065fdcd438918fd57a5708268b73fd4c00d6527e968fe8176680d04455381R79-R92) [[2]](diffhunk://#diff-309065fdcd438918fd57a5708268b73fd4c00d6527e968fe8176680d04455381L417-R500)

**Submission Checker Parameter Handling**
[...]

PR #617, opened 2026-07-01
fix(#616): label warmup runs in reportgen + remove --output-dir

### LINKED ISSUE #616: reportgen: label warmup run in stdout + remove --output-dir flag
## Context

Investigation of PR #613 (which proposed a per-folder rollup redesign to fix
apparent duplication in reportgen output) surfaced two real UX artifacts and
one design gap that should be addressed independently of that PR.

## Problem 1 — Warmup runs appear as "duplicates" in stdout

Training workloads produce **6 disk directories = 1 throwaway warmup + 5
submission runs** (by design). Only the 5 real runs are aggregated into
`results.json` / `results.csv` (dict deduplication via `run_id` handles that).

However, DLIO writes the warmup's `summary.start` value to match the first
real run's start time (not the warmup's own directory time), so the two
collide on `run_id`. The workload printer iterates the walker's pre-dedup
list, so the warmup **prints twice** in stdout while the JSON has it only
once. This is what led a reader to file #613 believing reportgen was
producing duplicated output.

Repro (from `sample_data/` Micron 9550_15TB v2.0 submission tree, cosmoflow):

| Directory (name → time)   | `summary.start` value           | Match?    |
|---------------------------|---------------------------------|-----------|
| `20250710_141219` WARMUP  | `2025-07-10T14:22:24.203210`    | mismatch — carries next dir's time |
| `20250710_142219` real 1  | `2025-07-10T14:22:24.203210`    | matches   |
| `20250710_143012` real 2  | `2025-07-10T14:30:16.920770`    | matches   |
| ... 3 more real runs      | (all self-consistent)           | matches   |

The DLIO-side stamp mismatch is not fixed here — we work around it in
reportgen by detecting collision-by-`run_id` and labelling the
lex-earlier-basename directory as the warmup.

### Proposed fix

- Detect warmup runs at accumulation time (colliding `run_id`, earlier
  `result_dir` basename).
- Label them in the workload printer as `[WARMUP, not aggregated]`.
- Add an inline comment near `_process_single_run` explaining the
  6=1+5 training / 1=1 checkpointing semantics and the DLIO
  collision-then-overwrite mechanism this relies on.

Example output:

\`\`\`
Training - resnet50, Accelerator: h100
    Runs:
      - training_run_resnet50_2025-07-10T19:56:24 [WARMUP, not aggregated]
      - training_run_resnet50_2025-07-10T20:06:28 [CLOSED]
      - training_run_resnet50_2025-07-10T20:29:12 [CLOSED]
      - training_run_resnet50_2025-07-10T20:51:47 [CLOSED]
      - training_run_resnet50_2025-07-10T21:14:30 [CLOSED]
      - training_run_resnet50_2025-07-10T21:37:07 [CLOSED]
\`\`\`

## Problem 2 — `--output-dir` allows separating summary from submission tree

`mlpstorage reports reportgen --output-dir /elsewhere` writes
`results.{csv,json}` outside the results directory hierarchy. This makes it
possible to submit a results tree that **does not include** the summary
files, which is not a supportable configuration — the summary is what the
MLCommons reviewer reads to see everything being submitted in one place.

### Proposed fix

Remove `--output-dir` entirely from the reportgen CLI (arg definition,
`HELP_MESSAGES` entry, static help text block). Hardcode the summary
location to `<results-dir>/results.{csv,json}`.

## Scope

Single PR covering:

1. Warmup detection + `[WARMUP, not aggregated]` label in stdout print.
2. Documentation comment near \`_process_single_run\` explaining the semantics.
3. Removal of \`--output-dir\` from CLI + help + \`report_generator.py\` write paths.
4. Unit tests for warmup labelling; inversion of \`test_output_dir_argument\`
   in \`tests/unit/test_cli.py\` to assert the flag is now rejected.

## Explicitly out of scope

Hardening the collision-drop mechanism itself (e.g., keying
\`self.run_results\` by \`result_dir\` instead of \`run_id\`, or explicitly
skipping the warmup rather than relying on \`os.walk\` lex-order overwrite).
Labelling in the walker UX addresses the visible problem without changing
the aggregate design.

## Related

- #613 (Wolfgang's proposed per-folder redesign; declined in favor of this)
- #606 (added \`--output-dir\` honoring in the first place; this PR removes it)
- #610 (independent reportgen fixes; keeps the one-aggregate design)

FileSystemGuy:
Closes #616.

## Summary

- Detect the DLIO warmup run in reportgen (colliding \`run_id\` from a
  mis-stamped \`summary.start\`) and label it as
  \`[WARMUP, not aggregated — dir: <basename>]\` in stdout instead of
  double-printing it as a submission run.
- Remove \`--output-dir\` from \`mlpstorage reports reportgen\`; hardcode
  \`results.{csv,json}\` to \`<results-dir>/\` so submitters can't
  accidentally exclude the summary from what MLCommons reviews.

## What the visible artifact looks like now

**Before** (six disk dirs = 1 warmup + 5 real runs for training; warmup
prints twice as CLOSED because its \`run_id\` collides with real run #1):
\`\`\`
    Runs:
      - training_run_resnet50_2025-07-10T14:22:24 [CLOSED]
      - training_run_resnet50_2025-07-10T14:22:24 [CLOSED]
      - training_run_resnet50_2025-07-10T14:30:16 [CLOSED]
      - training_run_resnet50_2025-07-10T14:38:09 [CLOSED]
      - training_run_resnet50_2025-07-10T14:46:02 [CLOSED]
      - training_run_resnet50_2025-07-10T14:53:55 [CLOSED]
\`\`\`

**After**:
\`\`\`
    Runs:
      - training_run_resnet50_2025-07-10T14:22:24 [WARMUP, not aggregated — dir: 20250710_141219]
      - training_run_resnet50_2025-07-10T14:22:24 [CLOSED]
      - training_run_resnet50_2025-07-10T14:30:16 [CLOSED]
      - training_run_resnet50_2025-07-10T14:38:09 [CLOSED]
      - training_run_resnet50_2025-07-10T14:46:02 [CLOSED]
      - training_run_resnet50_2025-07-10T14:53:55 [CLOSED]
\`\`\`

The aggregate \`results.{csv,json}\` are unchanged — the warmup was
already excluded via \`run_id\` dict dedup. This PR only fixes the
misleading stdout artifact + hardens the output location.

## Root cause of the collision

DLIO writes the warmup's \`summary.start\` value to match the FIRST real
run's start time (not the warmup's own directory timestamp). Detection:
on \`run_id\` collision at accumulation time, the run whose \`result_dir\`
basename is lex-earlier is the warmup. That basename is recorded in a
new \`ReportGenerator.warmup_result_dirs\` set and the workload printer
looks it up when rendering per-run lines. See the design comment on
\`_process_single_run\` for the full explanation and the training vs
checkpointing semantics.

## Why remove --output-dir

\`--output-dir\` (added in #606) let \`reports reportgen\` write the
summary anywhere on disk, which meant a submitter could technically
generate a submission tree that did not include the summary. That's not
a supportable configuration since the summary is what MLCommons reads
to see everything being submitted in one place. Hardcoded now to
\`<results-dir>/results.{csv,json}\`.

## Related

- #613 — Wolfgang's per-folder rollup redesign; declined in favor of
  this narrower fix (the design intent is one aggregate per submission).
- #606 — introduced \`--output-dir\` honoring; this PR removes it.
- #610 — independent reportgen fixes; keeps the one-aggregate design.

## Test plan

- [x] New file \`tests/unit/test_reportgen_warmup_labelling.py\` (9 tests):
  warmup collision detection (both walk orders), no-collision no-warmup,
  print-label appears with disk basename, print-order is lex(basename),
  no-warmup path unchanged, \`--output-dir\` args ignored, writes land in
  \`results_dir\`.
- [x] \`tests/unit/test_cli.py::test_output_dir_argument_rejected\` —
  inverted the prior test to assert \`--output-dir\` now raises
  \`SystemExit\` from argparse.
- [x] Full 4-suite sweep: \`tests/\` (2454 passed), \`mlpstorage_py/tests\`
  (780 passed, 1 xfailed), \`vdb_benchmark/tests\` (144 passed),
  \`kv_cache_benchmark/tests\` (238 passed, 2 deselected). Total 3,616
  passing, zero failures.
- [ ] Manual reportgen against a real training submission tree (e.g.
  the Micron 9550_15TB v2.0 sample) to confirm the WARMUP line renders
  as expected.


wolfgang-desalvador:
@FileSystemGuy this looks good, the only additional elements we need to port from my other branch is the following.

I think there is one last aspect we need to fix: the fact that in current implementation a single giant `result.json` in current implementation is written across models.

I fixed this in the [PR 620](https://github.com/mlcommons/storage/pull/620/changes) which is pointing to this branch 


| | Current branch | PR 620 |
|---|---|---|
| Where `results.{json,csv}` is written | One pair at `self.results_dir` (root), aggregating every model + every command | One pair per model folder, aggregating every command for that model |
| Grouping logic | None | New `_model_group_folder(result)` walks up from the run's `result_dir` leaf |
| Result for multi-model trees | All models concatenated into one file; per-model dirs empty | Each model's dir gets its own rollup |

Grouping walks up from the `<ts>/` leaf per benchmark type
(canonical layouts from `rules.utils.generate_output_location()`):

| Benchmark type    | Layout                                                              | Rollup lands at         |
|-------------------|---------------------------------------------------------------------|-------------------------|
| `training`        | `.../<system>/training/<model>/<command>/<ts>/`                     | `.../<model>/`          |
| `kv_cache`        | `.../<system>/kv_cache/<model>/<command>/<ts>/`                     | `.../<model>/`          |
| `checkpointing`   | `.../<system>/checkpointing/<model>/<ts>/` (no `<command>`)         | `.../<model>/`          |
| `vector_database` | `.../<system>/vector_database/<engine>/<index>/<command>/<ts>/`     | `.../<engine>/<index>/` |



---

wolfgang-desalvador (approved):
Please take a look at the attached PR before merging

Issue #681, opened 2026-07-05
[BUG]TypeError in run_data_matches_datasize (3.3.1): > not supported between int and str


xutao00090:
Problem

  Submission checker rule 3.3.1 (trainingRunDataMatchesDatasize) crashes with a TypeError when processing the retinanet training workload:
  TypeError: '>' not supported between instances of 'int' and 'str'
```
[ERROR] Exception occurred in /root/storage_results_v3/closed/xxx/results/xxxx/training/retinanet while running run_data_matches_datasize in TrainingCheck
Traceback (most recent call last):
  File "/crash/storage/mlpstorage_py/submission_checker/checks/base.py", line 67, in run_checks
    v = self.execute(check)
        ^^^^^^^^^^^^^^^^^^^
  File "/crash/storage/mlpstorage_py/submission_checker/checks/base.py", line 90, in execute
    return check()
           ^^^^^^^
  File "/crash/storage/mlpstorage_py/submission_checker/checks/training_checks.py", line 390, in run_data_matches_datasize
    and run_num_files_train > datagen_num_files_train):
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'int' and 'str'

```
training_20260705_171351_metadata.json:
```
"dataset": {
      "data_folder": "/mnt/xxx/retinanet",
      "format": "jpeg",
      "num_files_train": "68090280",
      "num_samples_per_file": 1,
      "record_length_bytes": 322957,
      "skip_listing": "True",
      "listing_validation_interval": "10000"
    }
```

  Root Cause

  At training_checks.py:390, the comparison:
  and run_num_files_train > datagen_num_files_train):
  compares two values read from JSON artifacts without type normalization:
  - run_num_files_train = summary.get("num_files_train") — from summary.json, can be either int or str depending on JSON serialization
  - datagen_num_files_train = params["dataset"].get("num_files_train") — from metadata.json / parameters, may survive YAML→JSON round-tripping as a string when the YAML source uses quoted scalars

  The same latent bug exists at line 370 (run_num_files_train < ds_num_files_train), though it may not have triggered yet.
  Affected Code
  mlpstorage_py/submission_checker/checks/training_checks.py, method run_data_matches_datasize, lines 368–370 and 388–390.
  Expected Behavior
  The comparison should coerce both operands to int before comparing, so a submission with mixed int/str JSON values does not crash the checker.

[...]

PR #684, opened 2026-07-05
fix(#681): coerce num_files_train to int in rule 3.3.1 before comparison

FileSystemGuy:
## Summary

- Fixes crash in rule 3.3.1 (`trainingRunDataMatchesDatasize`) when `num_files_train` is serialised as a quoted string in `summary.json` or metadata `parameters.dataset`
- Adds `_to_int()` helper and applies it at all three read sites so comparisons never see a mixed `int`/`str` pair

## Root cause

JSON/YAML round-tripping (observed on retinanet submissions with `"num_files_train": "68090280"`) produces string-typed integers. The `>` / `<` comparisons at lines 370 and 390 of `training_checks.py` did not coerce types, raising:

```
TypeError: '>' not supported between instances of 'int' and 'str'
```

## Fix

Added `TrainingCheck._to_int(v)` — returns `int(v)` or `None` on `None`/non-numeric input — and applied it at the three `num_files_train` read sites:

- `run_num_files_train` (run loop, `summary.json`)
- `num_files_train` in `_extract_latest_datagen_cardinality` (datagen metadata)
- `num_files_train` in `_group_datasize_by_data_dir` (datasize metadata)

## Test plan

New file `test_training_check_issue681.py`:
- Unit tests for `_to_int` covering int, str, None, float-string, garbage
- Integration tests for all mixed int/str combinations — no crash
- Regression test confirming `DATAGEN-OVERRUN` warning still fires after coercion

Parallel 4-suite sweep: 3754 passed, 1 skipped, 0 failures.

### DIFF
--- a/mlpstorage_py/submission_checker/checks/training_checks.py
+++ b/mlpstorage_py/submission_checker/checks/training_checks.py
@@ -354,7 +354,7 @@ class TrainingCheck(BaseCheck):
                 # 2.1.19 already flags missing summary; do not double-fire here.
                 continue
 
-            run_num_files_train = summary.get("num_files_train")
+            run_num_files_train = self._to_int(summary.get("num_files_train"))
             run_num_files_eval = summary.get("num_files_eval")
             run_data_dir = metadata.get("args", {}).get("data_dir")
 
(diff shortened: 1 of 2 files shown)


FileSystemGuy:
@fixes #684 681

Issue #682, opened 2026-07-05
Checkpoint write throughput regresses ~41% on POSIX/`local_fs` (mlpstorage 3.0.26 → 3.0.33) - Materially change of results


wolfgang-desalvador:
> **Severity:** High — regression landed **inside the submission window** and materially changes reported results.

## Summary

`llama3-8b` checkpoint **save** throughput dropped ~41% between mlpstorage
**3.0.26** and **3.0.33** on an identical cluster, workload, and DLIO checkpoint
code path (POSIX filesystem, `storage_type: local_fs`).

| Version | Commit | Checkpoint save I/O (GiB/s) |
|---|---|---|
| 3.0.26 | [`6576710`](https://github.com/mlcommons/storage/commit/6576710) | **~12.4** |
| 3.0.33 | [`75414e4`](https://github.com/mlcommons/storage/commit/75414e4) | **~7.4** |

Δ = **~−41%** on the same hardware and workload.

## Environment

- **Model / mode:** `llama3-8b`, checkpointing, `num_checkpoints_write=10`
- **Ranks:** `num_processes=8` (2 × 4), dual-socket 192 vCPU clients
- **Target:** POSIX filesystem — `storage_type: local_fs`, `o_direct: false`,
  `fsync: true`

## Root cause

AI Analysis below, which has been proven to be correct on real tests.

The regression is caused by **[PR #650](https://github.com/mlcommons/storage/pull/650) / fix([#642](https://github.com/mlcommons/storage/issues/642))**, which changed the
streaming-checkpoint **writer subprocess** context from `fork` to `forkserver`
in `mlpstorage_py/checkpointing/streaming_checkpoint.py`:

```python
# 3.0.26 (~12.4 GiB/s)
ctx = mp.get_context('fork')

# 3.0.33 (#642, ~7.4 GiB/s)
ctx = _writer_mp_context()   # prefers forkserver, falls back to spawn
```

#642 was a **correct** fix for a `fork`-after-live-Tokio deadlock **on the
object-store path** (the parent starts a Tokio runtime via
`ObjStoreLibStorage._preflight()` before forking). However it was applied
**unconditionally**, including the `backend='file'` / `local_fs` path, which
**never** starts a parent-side Tokio runtime and therefore never had the
deadlock. `forkserver` writers spawn from a clean daemon started before
MPI/NUMA placement, losing the parent's copy-on-write warm state and CPU/NUMA
locality with the shared-memory buffers, consistent with the ~41% drop.

[...]

Issue #689, opened 2026-07-06
Checkpoint write throughput is capped by CPU-side data generation, not storage (`dgen-py` under-threaded → Generation is the bottleneck)


wolfgang-desalvador:
> **Severity:** High — the reported checkpoint metric measures synthetic-data
> generation (CPU) throughput rather than storage throughput, so it under-reports
> the storage system and materially changes results.

## Summary

During a `llama3-70b` checkpoint **save** run, the streaming-checkpoint
**producer** (synthetic tensor generation via `dgen-py`) is a co-equal — and in
a share of chunks the *dominant* — bottleneck relative to storage I/O. The
producer is CPU-bound and under-threaded, so the reported checkpoint throughput
is limited by how fast the client can generate the checkpoint tensors, not by
how fast the storage target can absorb them.

Per-chunk profiling from a single run:

| Signal | Observed |
|---|---|
| Generator threads per rank | **3 threads** (`threads=3`) |
| Throughput ratio (gen/io) | **~0.9–1.1×** |
| Chunks where **Generation** is the bottleneck | **~35%** (1,327 of 3,840) |
| Pipeline overhead when Generation-bound | **up to ~10%** (vs ~3% when I/O-bound) |

Because generation and I/O run at a similar rate, the measured throughput sits
at whichever side is momentarily slower — and about a third of the time that is
the CPU-side generator, not the storage device.

## Environment

- **Model / mode:** `llama3-70b`, checkpointing, `num_checkpoints_write=10`
- **Ranks:** `num_processes=64` (16 × 4)
- **Client nodes:** 16 nodes, each with 192 vCPUs and 1.8 TiB of RAM, **reserved
  entirely for the job** (all cores available, ranks launched unbound — the
  3-thread generator is an application choice, not a scheduler/cgroup CPU limit)
- **Target:** POSIX filesystem — `backend: file`, `o_direct: false`
- **mlpstorage:** [`6576710`](https://github.com/mlcommons/storage/commit/6576710) (3.0.26)
- **Per-shard checkpoint size:** ~12.2 GB, streamed through a 4 × 32 MB buffer pool

## Analysis

The streaming checkpoint uses a producer–consumer pipeline: a producer
generates synthetic tensor data (`dgen-py`) into a small shared-memory buffer
pool, and a writer subprocess drains those buffers to storage. The intent is
that the writer (storage I/O) is the bottleneck so the run measures **storage**
throughput.

In practice the producer is initialised with only **3 worker threads**
[...]

PR #700, opened 2026-07-07
checkpoint: make writer start method selectable (fork/forkserver) with backend-aware default (#682)

### LINKED ISSUE #642: Streaming checkpoint writer deadlocks: multiprocessing `fork` after s3dlio's Tokio runtime is live
**Summary**
`StreamingCheckpointing.save()` starts its writer subprocess with `fork`. On the s3dlio backend the rank has already started s3dlio's Tokio runtime (via `ObjStoreLibStorage._preflight()` → `s3dlio.list()`), so `fork()` copies mutexes held by threads that don't exist in the child; the writer deadlocks the first time it re-enters s3dlio and the producer blocks on a full queue. Symptom: all ranks idle at ~0% CPU, writers stuck right after *"Attached to N buffers"*.

**Environment**
mlcommons/storage 3.0.28 @ [`76edfd3`](https://github.com/mlcommons/storage/commit/76edfd3) (latest `main`); Linux; object (s3dlio) backend.

**Affects**
`mlpstorage checkpointing run` **write** on **object storage** (`s3dlio`), any LLM model; observed during llama3 checkpointing S3 bring-up.

**Reproduce**
```bash
mlpstorage whatif checkpointing run object \
  --model llama3-405b --num-processes 512 --hosts worker-{0..31} \
  --client-host-memory-in-gb 128 \
  --checkpoint-folder s3://mlc-s3/checkpoints/llama3-405b/default \
  --results-dir /data/results --systemname "$SYSTEMNAME" \
  --num-checkpoints-write 10 --num-checkpoints-read 0 --allow-run-as-root
```

**Root cause**
In `save()`, the writer subprocess is created with `mp.get_context('fork')`. Forking after the Tokio runtime is live → futex deadlock in the child's first s3dlio call; the producer then blocks on a full `buffer_queue`.

**Proposed fix**
In [`mlpstorage_py/checkpointing/streaming_checkpoint.py`](https://github.com/mlcommons/storage/blob/main/mlpstorage_py/checkpointing/streaming_checkpoint.py), use `forkserver` (fallback `spawn`): it forks writers from a clean, early-started interpreter (no live Tokio) and, unlike `spawn`, doesn't re-import `__main__` (which caused MPI-rejoin hangs). IPC objects are already built from the same `ctx`.
```python
try:
    ctx = mp.get_context('forkserver')
except ValueError:
    ctx = mp.get_context('spawn')
```

zhenghh04:
## Summary

Fixes the ~41% checkpoint-save regression reported in #682 by making the
streaming-checkpoint **writer subprocess start method** a *selectable option*
with a **backend-aware default**, instead of the unconditional `forkserver`
introduced by #650 (fix #642).

### Background

- **#642** was a correct fix: on the **s3dlio object-storage** path the parent
  already has a live Tokio runtime (`ObjStoreLibStorage._preflight()` →
  `s3dlio.list()`) by the time the writer subprocess is created, so `fork()`
  copies that mutex state and the writer futex-deadlocks the first time it
  re-enters s3dlio.
- **#682**: #650 applied `forkserver` **unconditionally**, including the POSIX
  `file` / `local_fs` path, which **never** starts a parent-side Tokio runtime
  and never had the deadlock. There, `forkserver` loses the parent's
  copy-on-write warm state and CPU/NUMA locality with the shared-memory
  buffers, regressing `llama3-8b` checkpoint save ~41% (~12.4 → ~7.4 GiB/s) —
  and it landed inside the submission window.

### Change

The start method is now a choice with a backend-aware default:

| Target | `'auto'` default | Why |
|---|---|---|
| `backend='file'` / `local_fs` / scheme-less / `file://` | **`fork`** | no parent Tokio → safe + ~40% faster (#682) |
| `s3dlio`, `direct_fs`, `s3torchconnector`, `minio`, `s3://`/`az://`/`gs://`/`direct://` | **`forkserver`** | fork-after-live-Tokio deadlocks (#642) |

- Override via `mp_start_method=` constructor arg or
  `MLPS_CHECKPOINT_MP_START_METHOD` env var (constructor arg wins).
- Explicit `fork` on the object-storage path is **refused** (`ValueError`
  citing #642) rather than silently deadlocked.
- Platform-availability fallback preserved: `fork` → `forkserver` → `spawn`.
- `save()` now logs the resolved method, e.g.
  `[Main] Writer start method: fork (policy='auto', backend=file)`, so each run
  records which method produced its result.

### Implementation

- `_is_posix_file_backend()` mirrors `StorageWriterFactory`'s own file-vs-object
  resolution.
- `_resolve_mp_start_method()` — pure, platform-independent policy (returns
  `fork`/`forkserver`/`spawn` or raises).
- `_writer_mp_context(start_method, backend, uri_or_path)` — resolves policy,
  then degrades only for platform availability.
- New `mp_start_method` argument on `StreamingCheckpointing.__init__`.

### Tests

`tests/unit/test_streaming_checkpoint_mp_context.py` rewritten to lock **both**
properties:
- **#642 safety** — object storage never forks; `forkserver`-unavailable
  degrades to `spawn` (not `fork`); explicit `fork` on object storage raises.
- **#682 default** — the file backend forks by default (where `fork` exists);
  explicit override honored; invalid method raises.

```
16 passed
```

### Note for the WG (submission comparability)

This restores POSIX performance without reintroducing the object-store deadlock.
Because the headline metric depends on the method, a tagged release / rules
should still **pin the canonical value per `storage_type`** (the knob is for
override/debug); the `save()` log line makes the method auditable per run.

### DIFF
--- a/docs/Streaming-Chkpt-Guide.md
+++ b/docs/Streaming-Chkpt-Guide.md
@@ -212,6 +212,41 @@ checkpoint.save('s3://bucket/key', total_size)
 checkpoint.save('/nvme/checkpoint.pt', total_size)
 ```
 
+### Writer subprocess start method (`fork` vs `forkserver`)
+
+The streaming writer runs in a subprocess. The multiprocessing **start
+method** used to create it is selectable, with a **backend-aware default**:
+
+| Target | `'auto'` default | Rationale |
+|---|---|---|
+| POSIX file (`backend='file'`, or a scheme-less / `file://` path) | `fork` | The parent never starts s3dlio's Tokio runtime, so `fork` is safe **and** ~40% faster — it keeps the parent's copy-on-write warm state and CPU/NUMA locality with the shared-memory buffers ([#682](https://github.com/mlcommons/storage/issues/682)). |
+| Object storage (`s3dlio`, `direct_fs`, `s3torchconnector`, `minio`, or any `s3://`/`az://`/`gs://`/`direct://` URI) | `forkserver` | The parent has a live s3dlio Tokio runtime (via `ObjStoreLibStorage._preflight()`) by the time the writer is created; `fork()` copies that mutex state and the writer futex-deadlocks ([#642](https://github.com/mlcommons/storage/issues/642)). `forkserver` gets a clean interpreter without `spawn`'s `__main__` re-import (which hangs MPI ranks). |
+
+Leave `mp_start_method='auto'` (the default) and each backend gets the right
+method automatically. To override:
+
+```python
+# Constructor argument (highest precedence)
+checkpoint = StreamingCheckpointing(
+    backend='file',
+    mp_start_method='forkserver',   # 'auto' | 'fork' | 'forkserver' | 'spawn'
+)
+```
+
+```bash
+# Or via environment variable (constructor arg wins if both are set)
+export MLPS_CHECKPOINT_MP_START_METHOD=fork
+```
+
+Notes:
+- Requesting `mp_start_method='fork'` on an **object-storage** target raises a
+  `ValueError` (citing #642) rather than deadlocking.
+- If the platform lacks the requested method (e.g. no `fork`/`forkserver` on
+  Windows/macOS), it degrades gracefully: `fork` → `forkserver` → `spawn`.
+- `save()` logs the resolved method, e.g.
+  `[Main] Writer start method: fork (policy='auto', backend=file)`, so each run
+  records which method produced its result.
+
 ## Feature 4: Multi-Endpoint Load Balancing
 
 ### What It Does
--- a/mlpstorage_py/checkpointing/streaming_checkpoint.py
+++ b/mlpstorage_py/checkpointing/streaming_checkpoint.py
@@ -60,37 +60,111 @@ def _local_ranks_per_node(env=None):
     return 1
 
 
-def _writer_mp_context():
+# Selecting the streaming-checkpoint writer subprocess start method.
+#
+# The right method depends on the backend, so it is a *choice* with a
+# backend-aware default (issues #642 and #682):
+#
+#   * POSIX file backend (``backend='file'`` or a scheme-less / ``file://``
+#     path): the parent never starts s3dlio's Tokio runtime, so ``fork`` is
+#     safe *and* ~40% faster — it keeps the parent's copy-on-write warm state
+#     and CPU/NUMA locality with the shared-memory buffers (#682). Default:
+#     ``fork``.
+#   * Object-storage / s3dlio path (``s3dlio``, ``direct_fs``,
+#     ``s3torchconnector``, ``minio``, or any remote-scheme URI): the parent
+#     has already started s3dlio's Tokio runtime via
+#     ``ObjStoreLibStorage._preflight()`` → ``s3dlio.list()`` by the time the
+#     writer is created. ``fork()`` copies that mutex state without the
+#     threads holding the mutexes and the writer futex-deadlocks the first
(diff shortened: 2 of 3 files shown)


zhenghh04:
@russfellows @wolfgang-desalvador Please check whether this is OK. 

@FileSystemGuy , we should discuss how to handle this.

wolfgang-desalvador:
@zhenghh04 the changes at a "quick" look seem ok to me. Thank you for pulling this together so quickly

However I would like @FileSystemGuy to balance how we should handle this, since this would be changing again the behavior in the submission window.

russfellows:
Note, this does not ACTUALLY fix the bug in #689.  It helps address it, but is not the actual fix.  That said, the changes here propose by @zhenghh04  are good and helpful in mitigating bug 689.  My Claude just found the actual cause of 689, and it makes sense now that I see it.  So, I will approve and merge this PR, since it is helpful in enhancing the overall stability of the code.

russfellows (approved):
My claude reviewed this, and I pulled the PR and ran tests.  It does help mitigate the issues outlined.  No issues so I approve this.  However, it does not actually fix 689, so I will be submitting another PR for that fix.  But again, these changes by Huihuo are helpful, and I have tested them, so approving.

FileSystemGuy:
@russfellows @wolfgang-desalvador @zhenghh04 

Sorry for the delay in seeing this, my car had a flat tire that I needed to attend to.

While we're very late in the game, this PR *should*, if I understand correctly, narrowly change the behavior of one workload.  Since we know that its a dramatic change (one might say existential since it's that big), and everyone who re-runs this test will get a better score, we just need to know whether an org used the new code or the old one when they submit results.

FileSystemGuy (commented):
One operator-precedence bug in `_resolve_mp_start_method`:

```python
# line 83 — current (broken):
f"object-storage backend ({backend or uri_or_path!r}): ..."
# `!r` binds tighter than `or`, so this is `backend or (uri_or_path!r)`.
# When backend='s3dlio' the message prints  s3dlio  (no quotes).
# When uri_or_path is the fallback it prints 's3://...' (with quotes).

# fix:
f"object-storage backend ({(backend or uri_or_path)!r}): ..."
```

One character change — add parens around `backend or uri_or_path` so `!r` applies to the whole expression.

FileSystemGuy:
Reviewed the PR. The design is sound — clean separation of policy (`_resolve_mp_start_method`) from platform availability (`_mp_availability_fallback`), and the #642 guardrail (raising on `fork`+object-storage rather than silently deadlocking) is the right call. A few observations:

**One code defect — operator precedence in the error message**

```python
# line 83, current:
f"object-storage backend ({backend or uri_or_path!r}): ..."
```

`!r` binds tighter than `or`, so this is `backend or (uri_or_path!r)`. When `backend='s3dlio'` the message prints `s3dlio` without quotes; only when `uri_or_path` is the fallback does `!r` fire. Fix:

```python
f"object-storage backend ({(backend or uri_or_path)!r}): ..."
```

**Minor: `ValueError` for an invalid `mp_start_method` string is deferred to `save()` time**

The policy string (`"threads"`, etc.) is stored at `__init__` but not validated until `save()` calls `_resolve_mp_start_method`. Could validate at construction since the method string is fully known then — a constructor that accepts a bad argument and blows up later is harder to debug at scale.

**Submission comparability note**

The `save()` log line (`[Main] Writer start method: fork (policy='auto', backend=file)`) makes each run auditable, which is good. Worth confirming with the WG whether existing POSIX-file submissions (made under the unconditional `forkserver` of #650) need to be re-run or flagged — the recovered ~40% on the file path is a material result difference.

PR #732, opened 2026-07-09
feat(#731): validate training datagen hierarchy and emit self-describing manifest

### LINKED ISSUE #731: Training datagen: validate output hierarchy and emit self-describing manifest
## Problem

`mlpstorage training datagen` currently produces:

- A results-dir leaf at `<results-dir>/.../training/<model>/datagen/<ts>/` with `training_datagen.stdout.log`, `training_datagen.stderr.log`, `dlio.log`, `training_<ts>_metadata.json`, and `dlio_config/` (containing `config.yaml`, `hydra.yaml`, `overrides.yaml`).
- The generated dataset under `<data-dir>/<model>/`.

Two related gaps:

1. **No completeness check on the results-dir leaf** at either the moment datagen finishes (operator-facing) or at `reports reportgen` time (submission audit). Rule §2.1.14 / §2.1.15 in `mlpstorage_py.submission_checker` validate these files, but only when someone runs the standalone submission checker tool — the operator running `mlpstorage training datagen` gets no immediate feedback.

2. **No self-describing manifest for the generated dataset.** There is no on-disk record that says "this `<data-dir>/<model>` tree was generated with num_files_train=X, num_samples_per_file=Y, record_length_bytes=Z". A future `mlpstorage training run` command has no way to verify that its requested workload volume is compatible with the dataset it is about to consume.

## Scope of this issue

**In:**

- New helper module `mlpstorage_py/rules/datagen_hierarchy.py` with four functions:
  - `validate_supported_model(model, mode)` — mode-aware allowlist enforcement (closed/open = `MODELS_CLOSED`/`MODELS_OPEN`; whatif skips).
  - `assert_data_dir_hierarchy_absent(data_dir, model)` — refuse to overwrite a non-empty `<data-dir>/<model>` tree. Rationale: re-datagen with different `num_samples_per_file` (or any per-file-byte-size knob) would leave old and new files interleaved on filename collision, silently corrupting the dataset.
  - `validate_datagen_leaf(leaf_path)` — stat-only presence check reusing `DATAGEN_REQUIRED_FILES` / `DATAGEN_REQUIRED_FOLDERS` from `submission_checker/constants.py`. Never raises; returns a list of missing-item strings.
  - `write_datagen_manifest(...)` — write `<data-dir>/<model>/.mlps-datagen-manifest.json` with the three DLIO knobs plus provenance. Loud failure if any of the three fields is absent from the merged config.

- Pre-run guards wired into `TrainingBenchmark.__init__` (before `add_datadir_param`): supported-model check + refuse-to-overwrite. Whatif skips both. Local filesystem only (object storage is out of scope for the refuse-to-overwrite guard).

- Post-run actions wired into `TrainingBenchmark._run` (after DLIO datagen succeeds): leaf-presence WARN + manifest write. Local filesystem only.

- Reportgen datagen-side validation in `report_generator._process_workload_groups`: supported-model INVALID gate + leaf-presence WARN in the row's issues column. Whatif skips both (matches D-29 policy). No `--data-dir` access; no tree walks anywhere.

**Out:**

- Any `--data-dir` tree walk to verify file counts, sizes, or subfolder structure. `--data-dir` can be extremely large (object-bucket-backed data-dirs may take minutes to walk) — validation trusts DLIO's own file-count invariants.
- Object-storage manifest write. S3 presence semantics need a different API and the future run-vs-datagen check will need a separate discovery path.
- A run-side reader of the manifest (future work — see below).

## Design decisions worth preserving

| Decision | Why |
|---|---|
| Manifest at `<data-dir>/<model>/.mlps-datagen-manifest.json` (per-model nested, not at data-dir root) | One `--data-dir` may hold multiple model datasets. |
| Manifest fields kept minimal (three DLIO knobs + provenance) | The future run-vs-datagen check needs only these three; adding more risks drift. |
| Refuse to overwrite existing `<data-dir>/<model>` | Silent partial overwrite via filename collision would corrupt performance measurements. |
| Whatif skips all new checks | Matches the D-29 whatif-exemption policy already in reportgen (issue #717) and the user direction that whatif carries no significant validation. |
| Supported models: `unet3d`, `retinanet` (`MODELS_CLOSED` == `MODELS_OPEN` in v3.0) | Rules.md v3.0. cosmoflow / resnet50 are v2.0-only. |
| WARN messages prefixed `[WARN]` in the reportgen `issues` column | Distinguishes from `[INVALID]`-triggering D-24 templates for CSV consumers. |
| Manifest write failure returns `EXIT_CODE.FAILURE` | A workload YAML lacking the three DLIO fields is a real bug; silent zero-valued writes would corrupt the future run-vs-datagen check. |

## Future work (informed by this scope but explicitly separate)

`mlpstorage training run` should read `<data-dir>/<model>/.mlps-datagen-manifest.json` and refuse (or WARN) if the run's requested `num_files_train`, `num_samples_per_file`, or effective byte volume exceeds what the dataset supports. Locking the manifest schema now with the tightest set of fields (three knobs + provenance) keeps that future PR narrow.

FileSystemGuy:
Closes #731.

## Summary
- Adds `mlpstorage_py/rules/datagen_hierarchy.py` with four helpers: mode-aware supported-model check, refuse-to-overwrite guard on `<data-dir>/<model>`, stat-only leaf-presence check, and manifest writer.
- Wires pre-run guards into `TrainingBenchmark.__init__` (fire before any directory is created) and post-run leaf WARN + manifest write into `TrainingBenchmark._run` (fire after DLIO succeeds).
- Wires supported-model INVALID gate + leaf-presence WARN into `report_generator._process_workload_groups` for training datagen groups. Whatif skips all new checks (matches D-29 policy from #728).
- No `--data-dir` tree walk anywhere. `--data-dir` may be extremely large; validation trusts DLIO's own file-count invariants.
- Manifest at `<data-dir>/<model>/.mlps-datagen-manifest.json`: schema_version, model, num_files_train, num_samples_per_file, record_length_bytes, created_at (ISO 8601 UTC), mlpstorage_version, source_datagen_result_dir. Future PR will read this from `mlpstorage training run` to verify the requested workload fits the dataset.

## Test plan

**RED-first TDD across four pairs of commits:**

- [x] `test_datagen_hierarchy.py` (41 unit tests) — pure helper behavior: allowlist per mode, refuse-to-overwrite states, leaf-presence returns, manifest fields + missing-field loud-failure.
- [x] `test_datagen_hierarchy_wiring.py` (13 integration tests) — TrainingBenchmark pre-run guards fire only for datagen non-whatif; post-run manifest write, leaf WARN, and object-storage/whatif skips.
- [x] `TestDatagenReportgenValidation` in `test_aggregation.py` (5 tests) — reportgen INVALID on unsupported model, WARN on missing leaf files, whatif exemption, and healthy-leaf baseline.

**Full CI parity locally (matrix from `.github/workflows/test.yml`):**

- [x] `uv run pytest tests` — 2806 passed, 1 skipped, 13 deselected
- [x] `uv run pytest mlpstorage_py/tests` — 834 passed
- [x] `uv run pytest vdb_benchmark/tests` — 174 passed
- [x] `uv run pytest kv_cache_benchmark/tests` — 238 passed

## Notes

- No changelog entry added — recent fixes (#728, #729) landed on `main` without touching `BugsFixedInVersions.md`; the maintainer bundles docs updates on version bump.
- Not in scope, called out in issue #731 for future work: `mlpstorage training run` reading the manifest to compare its requested workload size against what the dataset supports.

### DIFF
--- a/mlpstorage_py/benchmarks/dlio.py
+++ b/mlpstorage_py/benchmarks/dlio.py
@@ -16,7 +16,9 @@ from mlpstorage_py.errors import ConfigurationError, ErrorCode
 from mlpstorage_py.rules import calculate_training_data_size, HostInfo, HostMemoryInfo, HostCPUInfo, ClusterInformation
 from mlpstorage_py.rules.datagen_hierarchy import (
     assert_data_dir_hierarchy_absent,
+    validate_datagen_leaf,
     validate_supported_model,
+    write_datagen_manifest,
 )
 from mlpstorage_py.utils import (read_config_from_file, create_nested_dict, update_nested_dict, generate_mpi_prefix_cmd)
 from mlpstorage_py.storage_config import resolve_object_storage_config
@@ -957,7 +959,49 @@ class TrainingBenchmark(DLIOBenchmark):
         except Exception as e:
             self.logger.error(f'Error occurred while executing command: {str(e)}')
             return EXIT_CODE.FAILURE
+        # Post-datagen actions: leaf WARN + self-describing manifest
+        # write. Runs only after DLIO's datagen command returned
+        # success and only outside whatif (whatif is simulation and
+        # skips per D-29 policy).
+        if self.args.command == "datagen" and self.args.mode != "whatif":
+            try:
+                self._post_datagen_actions()
+            except ConfigurationError as e:
+                # A workload YAML that lacks the three DLIO fields is
+                # a real bug — surface loudly rather than write a
+                # silently-broken manifest.
+                self.logger.error(
+                    f'Post-datagen validation failed: {e}'
+                )
+                return EXIT_CODE.FAILURE
         return EXIT_CODE.SUCCESS
+
+    def _post_datagen_actions(self):
+        """WARN for missing leaf files, then write the datagen manifest.
+
+        Leaf-presence check runs unconditionally — any missing file
+        surfaces as a WARN but does not flip the exit code. Manifest
+        write is local-storage-only (S3 presence semantics are a
+        separate API and the future run-vs-datagen size check would
+        need a different discovery path for object storage).
+        """
+        missing = validate_datagen_leaf(self.run_result_output)
+        for item in missing:
+            self.logger.warning(
+                f'Datagen output leaf incomplete: {item}'
+            )
+
+        storage_type = self.params_dict.get("storage.storage_type", "local")
+        if storage_type != "local":
+            return
+        dataset_params = (self.combined_params or {}).get("dataset", {})
+        manifest_path = write_datagen_manifest(
+            data_dir=self.args.data_dir,
+            model=self.args.model,
+            dataset_params=dataset_params,
+            source_datagen_result_dir=self.run_result_output,
+        )
+        self.logger.info(f'Datagen manifest written: {manifest_path}')
 
 
 class CheckpointingBenchmark(DLIOBenchmark):
--- a/mlpstorage_py/benchmarks/dlio.py
+++ b/mlpstorage_py/benchmarks/dlio.py
@@ -677,18 +677,14 @@ class TrainingBenchmark(DLIOBenchmark):
             # directories are created if we abort. Whatif skips both
             # guards (matches D-29 reportgen policy and the "no
             # significant validation" direction for whatif).
+            # The refuse-to-overwrite helper detects object-storage
+            # URIs via scheme and dispatches to s3dlio internally, so
+            # no storage_type gate is needed here.
             if self.args.command == "datagen" and self.args.mode != "whatif":
                 validate_supported_model(self.args.model, self.args.mode)
-                storage_type = self.params_dict.get(
-                    "storage.storage_type", "local"
+                assert_data_dir_hierarchy_absent(
+                    self.args.data_dir, self.args.model
                 )
-                if storage_type == "local":
-                    # Object-storage (S3) presence checks are a
-                    # different API and are out of scope for the
-                    # local-filesystem refuse-to-overwrite guard.
-                    assert_data_dir_hierarchy_absent(
-                        self.args.data_dir, self.args.model
-                    )
             # The datasize command uses --data-dir and needs to generate a command that also calls --data-dir
             # The add_datadir_param would convert --data-dir to --dataset.data_folder which is invalid to
             # mlpstorage.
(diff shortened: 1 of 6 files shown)


FileSystemGuy:
Extended per discussion — now covers object-storage parity for `--data-dir`.

## What's added
- URI-aware dispatch in `mlpstorage_py/rules/datagen_hierarchy.py`. New `_is_object_uri` helper detects `s3://`, `gs://`, `az://`, `direct://` (schemes s3dlio abstracts uniformly). Local paths and `file://` stay in the existing branch.
- `assert_data_dir_hierarchy_absent` dispatches to `s3dlio.list(uri, recursive=False)` for object storage — a delimiter-based listing that returns only direct children, bounded regardless of how many objects sit under a prior datagen's prefix. Any s3dlio exception → `ConfigurationError` (fail-safe abort per design).
- `write_datagen_manifest` dispatches to `s3dlio.put_bytes` for object storage. Manifest body is byte-identical across backends so the future run-side reader is one code path.
- The two `storage_type == 'local'` gates in `TrainingBenchmark` are removed — the URI-aware helpers own backend dispatch now.

## Test coverage
- **26 new unit tests** in `TestAssertDataDirHierarchyAbsentObjectStorage` and `TestWriteDatagenManifestObjectStorage`, parametrized across five URI-scheme fixtures. `s3dlio` is patched at the module boundary — no live endpoint required.
- **3 new integration tests** in `TestTrainingBenchmarkObjectStorageWiring` proving `TrainingBenchmark.__init__` invokes `s3dlio.list` for the refuse guard, `TrainingBenchmark._run` invokes `s3dlio.put_bytes` for the manifest write, whatif skips both.

## Rationale for depending on s3dlio's upstream tests
Discussed in the PR review: s3dlio is already the object-storage backend for the checkpointing benchmark's production I/O (GB-sized model checkpoints). Our sentinel operation is a bounded `LIST` and a ~1 KB `PUT` — orders of magnitude simpler than what s3dlio already carries in this codebase. Our tests cover URI parsing, correct s3dlio call arguments, empty/non-empty LIST semantics, and fail-safe abort on any s3dlio exception. What we don't test is wire-level correctness against a real endpoint, which is the trust boundary the checkpointing benchmark already crosses.

## Full CI parity (post-merge with `main`)
- [x] `uv run pytest tests` — 2835 passed, 1 skipped
- [x] `uv run pytest mlpstorage_py/tests` — 834 passed
- [x] `uv run pytest vdb_benchmark/tests` — 174 passed
- [x] `uv run pytest kv_cache_benchmark/tests` — 238 passed

The future `mlpstorage training run` command's reader can now use the same `_is_object_uri` + s3dlio dispatch pattern, keeping that follow-up PR narrow.

Issue #724, opened 2026-07-08
3.0.37 code-image pool: runs create the pool image but no `.mlps-image-pool` sentinel / `.mlps-code-image` pointers — how do we complete the migration?


gaikwadabhishek:
On 3.0.37 (content-addressed code-image pool, #710), `mlpstorage validate` fails on the
pool structure and I can't find the intended way to fix it.

### What we did

Ran the checkpointing suite on 3.0.37 into an `init`'d results dir, then `mlpstorage validate`.

### What happens

The run **created the pool image** — `\<results-dir\>/nvidia/code-1312df84/` with a
`.code-hash.json` (`algorithm: md5-tree-v2`, the full hash) — but there's **no
`nvidia/.mlps-image-pool` sentinel and no per-run-leaf `.mlps-code-image` pointers**. So
validate fails:

```
[CHECK-01 poolPointerResolution] .../nvidia: No pool root found for org nvidia:
    missing .../nvidia/.mlps-image-pool. Run mlpstorage to migrate.
[CHECK-04 poolLegacyCheck] .../nvidia: Partial migration detected for org nvidia
    (pool images found but .mlps-image-pool sentinel absent). Run mlpstorage to complete migration.
[2.1.2 topLevelSubdirectories] .../nvidia: unexpected top-level directory 'nvidia'
    (expected only 'closed'/'open', or a recognized pool root containing .mlps-image-pool)
```

So the pool is half-built: image present, sentinel + pointers absent.

### What we tried

- **`mlpstorage init nvidia <dir>`** — runs clean but does **not** create the
  `.mlps-image-pool` sentinel or the per-leaf pointers.
- **`reports reportgen`** — no `--output-dir` option in this version (`unrecognized arguments`).
- Looked for a `migrate` subcommand — the top-level help lists only
  `closed | open | whatif | init | reports | history | lockfile | version | validate | rules-coverage`.

The errors say "Run mlpstorage to migrate" / "to complete migration," but we can't find
which command does that.

### The question

What's the intended flow to complete the pool migration — i.e. to write the
`nvidia/.mlps-image-pool` sentinel and the per-run-leaf `.mlps-code-image` pointers?

- Is a benchmark `run` supposed to write them automatically (and it's a bug that our runs
  only produced the pool image)?
- Or is there a specific command/flag we're missing?
- Or is a legacy `code/`-layout tree expected, with the pool auto-created on validate?

(Also hitting the mirror of this from the other side: a **3.0.36** submission with the legacy
[...]

PR #734, opened 2026-07-09
fix(#724): make pool-check migration messages actionable

FileSystemGuy:
## Summary

Addresses the UX gap in #724.

The reporter's underlying failure modes (missing `.mlps-image-pool` sentinel, missing per-leaf `.mlps-code-image` pointers) are already fixed on main by PR #727 (auto-write sentinel) and PR #729 (pointer/reserve leaf collision). But the pool-check error messages that surface those states still say "Run mlpstorage to migrate" — a verb that doesn't exist. Auto-heal fires inside `_check_and_migrate_legacy_layout` at `mlpstorage_py/main.py:230`, which runs only on submission-mode invocations (never during `mlpstorage validate`). Users hit a dead end trying to find a migrate subcommand.

- Rewrite the three messages to name the exact form that triggers auto-heal: `` `mlpstorage {closed|open} <benchmark> {datasize|datagen|run}` ``.
- Add the honesty line: `` `mlpstorage validate` does not migrate `` — so users don't wait for validate to fix its own errors.
- Affected: CHECK-01 poolPointerResolution (no pool root), CHECK-04 poolLegacyCheck (legacy `code/` layout), CHECK-04 poolLegacyCheck (partial migration / D-91).
- TDD: commit 1 is RED (three tests pinning the new substrings, all failing on unchanged code); commit 2 is GREEN (production message rewrite).

The reporter should also upgrade past `adf72cd` and re-run any submission-mode command to retro-heal the sentinel; existing pointer-less result leaves still require re-running the affected benchmark (there is no repair tool for those).

## Test plan

- [x] `uv run pytest tests/` — 2751 passed, 1 skipped, 13 deselected
- [x] `uv run pytest mlpstorage_py/tests/` — 834 passed
- [x] `uv run pytest vdb_benchmark/tests/` — 174 passed
- [x] `uv run pytest kv_cache_benchmark/tests/` — 238 passed
- [x] Three new RED tests fail on unchanged code, pass with the fix
- [x] Existing pool-structure tests unchanged (`Partial migration detected` / `Legacy code/ layout detected` substrings preserved)
- [ ] Reporter confirms the new messages point them to a working recovery flow

### DIFF
--- a/mlpstorage_py/submission_checker/checks/pool_structure_checks.py
+++ b/mlpstorage_py/submission_checker/checks/pool_structure_checks.py
@@ -180,7 +180,9 @@ class PoolStructureCheck(BaseCheck):
                         "CHECK-01", "poolPointerResolution",
                         str(org_root),
                         "No pool root found for org %s: missing %s/.mlps-image-pool. "
-                        "Run mlpstorage to migrate.",
+                        "Auto-migration fires on the next `mlpstorage {closed|open} "
+                        "<benchmark> {datasize|datagen|run}` invocation against this "
+                        "--results-dir; `mlpstorage validate` does not migrate.",
                         submitter, str(org_root),
                     )
                     valid = False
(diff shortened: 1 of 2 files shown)


FileSystemGuy:
Fixes issue #724

Issue #489, opened 2026-06-22
vdb_benchmark: simple_bench can silently continue with empty FLAT ground-truth collection after Milvus copy fallback fails


lou-lydiksen-purestorage:
vdb_benchmark simple_bench: FLAT GT copy fallback uses id > -2**63, then reports success with 0 copied vectors

VectorDB run file: recall can be invalid because simple_bench does not abort when _flat_gt is empty

simple_bench FLAT ground-truth creation fails on Milvus int64 cursor and produces recall.num_queries_evaluated=0

========

This worked:

```
./mlpstorage closed vectordb datagen file \
--host 127.0.0.1 \
--port 19530 \
--config default \
--collection mlps_1536dimension_1000000vectors_16shards__distributionuniform_batchsize1000_chunksize10000_indextypeDISKANN \
--dimension 1536 \
--num-vectors 1000000 \
--num-shards 16 \
--distribution uniform \
--batch-size 1000 \
--chunk-size 10000 \
--index-type DISKANN \
--force \
--results-dir /mnt/mlperf/$HOSTNAME/vdb_results \
--debug 2>&1 | tee "/home/ir/mlperf/OUTPUT/mlpstorage_closed_vectordb_datagen_file_1536dimension_1000000vectors_16shards__distributionuniform_batchsize1000_chunksize10000_indextypeDISKANN_`getnow`.out"
```

========

This did not work:

```
./mlpstorage closed vectordb run file \
--host 127.0.0.1 \
--port 19530 \
--config default \
--collection mlps_1536dimension_1000000vectors_16shards__distributionuniform_batchsize1000_chunksize10000_indextypeDISKANN \
--vector-dim 1536 \
--num-query-vectors 1000000 \
--num-query-processes 16 \
--batch-size 1000 \
--vdb-index DISKANN \
--results-dir /mnt/mlperf/$HOSTNAME/vdb_results \
--debug 2>&1 | tee "/home/ir/mlperf/OUTPUT/mlpstorage_closed_vectordb_run_file_1536dimension_1000000vectors_16shards__distributionuniform_batchsize1000_chunksize10000_indextypeDISKANN_`getnow`.out"
```

========

ChatGPT analysis:

Here is a GitHub-ready issue summary you can paste or adapt.

---

## Summary

`mlpstorage closed vectordb run file` can silently produce invalid VectorDB benchmark results when FLAT ground-truth collection creation fails in `simple_bench.py`.

[...]

Issue #625, opened 2026-07-01
[VDB] recall is not meaningful because the dataset is uniformly random vectors


ByoungjunSeo:
recall is computed over uniformly random vectors, so the values come out too low to be meaningful (random high-dimensional vectors have no real neighbor structure). We understand from WG discussion that this is a recognized tension (realistic data vs. its CPU/memory cost), and that #489 only added a guard for empty / under-covered ground truth without changing data realism. Filing/commenting mainly to confirm current status and whether any change is planned for v3.0. (If a discussion already tracks this, please point us to it and we'll move the comment there.)

FileSystemGuy:
@idevasena

idevasena:
@ByoungjunSeo Thanks for filing this.

The ground truth itself is exact and correct. GroundTruthBuilder does streaming brute-force top-K over every block, and the flat_index truth mode queries a FLAT companion collection. So recall@k is mathematically well-defined even on random data. The problem is entirely in what that number discriminates.
Root cause is in `generator.py::_generate_block`. The default is `rng.random((n, d)) — i.i.d. uniform [0,1)` — then L2-normalized. Two consequences at d=1536:

Uniform [0,1) puts every vector in the positive orthant, so after normalization all vectors cluster around the diagonal direction. I measured (50K db vectors, 1536-dim, seeds 42/99 matching the code): mean pairwise cosine similarity 0.7498 with std 0.008. The true nearest neighbor sits at ~0.780 and the 100th at ~0.770. Relative contrast (mean distance / NN distance) is 1.135 — essentially no contrast. The normal distribution option is only marginally better (RC ≈ 1.12).

The gap between the 100th and 101st neighbor averages ~2×10⁻⁵ in similarity. That's within the noise floor of accumulated float32 error across a 1536-element dot product, so the boundary of the "true" top-100 set is effectively arbitrary — NumPy BLAS, Milvus's internal FLAT scan, and DiskANN's distance kernel can legitimately disagree on membership. This is the same fragility that made the #489 empty/under-covered GT guard necessary; #489 treated a symptom of this root cause.

[...]

PR #789, opened 2026-07-14
Fix(vdb_benchmark#625): planted query mode and tie-aware epsilon recall

idevasena:
Recall over i.i.d. uniform vectors is non-discriminative: at 1536-d the query-to-corpus cosine similarity concentrates (relative contrast ~1.1) and the top-k boundary gap falls within float32 matmul noise, so exact set-intersection recall penalizes numerically-tied neighbors and does not transfer across index configurations.

Option 1 - query_mode=planted: queries are perturbations of stored database vectors (bit-exact block-0 replay in the orchestrator; live pk-sampled fetch in simple_bench/enhanced_bench), planting a genuine near neighbor per query. Restores relative contrast from ~1.1 to >100 with zero change to ingest, index build, or storage footprint.

Option 2 - recall_epsilon: tie-aware recall credits returned neighbors whose ground-truth score is within epsilon of the k-th neighbor's. Exact recall is always reported alongside. GT similarities are carried through GroundTruthBuilder / ground_truth.npz / hit.distance capture; the backend search API is unchanged.

Defaults (query_mode=independent, recall_epsilon=0) preserve historical behavior bit-for-bit. flat_index truth mode falls back to exact recall with a warning. Verdict semantics from #489/#572 are unchanged.

### DIFF
--- a/vdb_benchmark/README.md
+++ b/vdb_benchmark/README.md
@@ -51,6 +51,9 @@ The `mlpstorage` path is recommended for standard benchmark workflows.
 - [8. End-to-End Examples](#8-end-to-end-examples)
 - [9. Enhanced Benchmark Full Reference](#9-enhanced-benchmark-full-reference)
 - [10. Metrics and Measurement](#10-metrics-and-measurement)
+  - [Recall Measurement](#recall-measurement)
+  - [Query Modes and Recall Semantics (Issue #625)](#query-modes-and-recall-semantics-issue-625)
+  - [Guidance for Submitters: Comparability and Reruns](#guidance-for-submitters-comparability-and-reruns)
 - [11. Testing and Validation](#11-testing-and-validation)
 - [12. Troubleshooting](#12-troubleshooting)
 - [13. Contributing](#13-contributing)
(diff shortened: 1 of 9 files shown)


FileSystemGuy:
@idevasena Is this a new workload or a correctness-fix to an existing workload or a bug-fix or ...?

I ask because we need to provide guidance on whether submitters need to rerun or not, etc.

idevasena:
@FileSystemGuy It's a recall metric definition enhancement change given the uniform random vector dataset. Pre-this PR the workload never computed wrong numbers — ground truth was exact, recall@k was correct. What #625 identified is that the correctly-computed number has almost no discriminative power on this data distribution — a methodology weakness.

Default path is still the current independent query mode. 

**Existing and in-flight submissions:** no rerun, no restatement. Defaults (query_mode: independent, recall_epsilon: 0) are identical to current behavior — verified by test down to the query RNG stream. The PR can merge mid-round without touching anyone.

**Current round: available as a diagnostic mode only.** Submitters may run planted/ε for their own analysis, but official results stay on the existing definition.

I kept the PR in draft state. If you're ok with merging, I can make it ready for review.

FileSystemGuy:
@idevasena Sounds great to me!

FileSystemGuy:
@idevasena I would merge this is right now but I wasn't sure if you were completely done with it or not.  Let me know if it's ready or you're still working on it.  Thanks!

idevasena:
@FileSystemGuy its ready for review and merge now. Thank you!

Issue #696, opened 2026-07-07
§4.7.1: mandated write/read split can't meet its own ≤30s gap (read-invocation startup ~46s); submission_checker accepts combined mode — please reconcile


Victor-SongZhh:
## Summary

For a CLOSED `llama3-8b` checkpointing run where the per-client checkpoint size is below the 3× memory threshold, **Rules.md §4.7.1 prose and the `submission_checker` implementation disagree on what is required**, and the prose-mandated path appears physically impossible on a normal setup:

1. **Prose (`Rules.md` §4.7.1):** *"When a flush is required, the submitter **must** execute the run in two invocations… The validator **must confirm this split occurred** and that the inter-phase gap did not exceed 30 seconds."*
2. **Implementation (`submission_checker`, 3.0.34):**
   - `checkpoint_data_size_ratio` (rule 4.3.1) emits a **`warn_violation`** ("advisory … not a hard violation") when checkpoint-per-node < 3× memory — it does **not** force a split.
   - `cache_flush_validation` (rule 4.7.1) does `if not pairs: return valid` — a **combined single-invocation** (no split pair) **passes with no violation** ("Combined-mode submissions … are valid per Pitfall 15").
   - `checkpoint_invocation_structure` (rule 4.7.1) explicitly accepts **single-invocation (10/10) OR** the two-invocation split.
3. **The mandated split is unmeetable:** when we run the split as §4.7.1 prescribes, the inter-phase gap is **~55s**, and the flush itself is only ~1s of that — the remaining ~46s is the **read-invocation's own startup** (framework import + CAP env-validation + MPI spawn) before it records `start_time`. Since `cache_flush_validation` measures `read.start_time − write.end_time` against a strict 30s bound, the split **hard-fails 4.7.1** on a setup where nothing is actually wrong.

Net: the prose demands a split that can't pass its own ≤30s check, while the checker accepts combined mode. A submitter following the prose fails; a submitter following the checker passes.

## Evidence (single client, 8× processes, `llama3-8b`, `storage_type: local_fs` / a FUSE-backed POSIX filesystem)

| Form | save GiB/s | load GiB/s | timestamps | 4.7.1 gap check |
|---|---|---|---|---|
| Two-invocation split (10/0 → flush → 0/10) | 4.28 | 5.68 | 2 | **FAIL — gap 54.6s > 30s** |
| Combined single invocation (10/10) | 4.32 | 5.65 | 1 | PASS (no split pair) |

[...]

PR #824, opened 2026-07-21
Fix calculation of checkpoint gap in reports

wolfgang-desalvador:
This pull request updates the handling of benchmark invocation timing metadata to improve the accuracy of cache-flush gap measurement and ensure compliance with the Rules.md specification. The main changes introduce new fields for invocation bookends, propagate them throughout the data model, and update logic in the checkpointing submission checker to prefer these authoritative timestamps.

**Enhancements to timing metadata handling:**

* Added `invocation_start_time` and `invocation_end_time` fields to the `BenchmarkRunData` dataclass, including documentation explaining their significance for cache-flush gap measurement and their use as authoritative origins per Rules.md §4.7.1.
* Updated the `parse` and `_from_metadata` methods in `models.py` to extract and populate the new invocation bookend fields from metadata, falling back to empty strings if not present. [[1]](diffhunk://#diff-456390226f02370ad924ea8673de4bc509f7e3c7b2b6f108be7a26776dc4d750R854-R855) [[2]](diffhunk://#diff-456390226f02370ad924ea8673de4bc509f7e3c7b2b6f108be7a26776dc4d750R977-R978) [[3]](diffhunk://#diff-456390226f02370ad924ea8673de4bc509f7e3c7b2b6f108be7a26776dc4d750R840)
* Added property accessors for `invocation_start_time` and `invocation_end_time` in the model class for convenient and consistent access.

**Submission checker logic improvements:**

* Modified the checkpointing submission checker to prefer the new invocation bookend fields (`invocation_end_time` and `invocation_start_time`) when calculating the cache-flush gap, with a fallback to the older summary fields for backward compatibility. This ensures startup/collection overhead is not incorrectly included in gap measurements.
* Updated error reporting in the checker to display the correct fields used for timing, reflecting the new logic.

### DIFF
--- a/mlpstorage_py/rules/models.py
+++ b/mlpstorage_py/rules/models.py
@@ -84,6 +84,15 @@ class BenchmarkRunData:
     result_dir: Optional[str] = None
     accelerator: Optional[str] = None
     end_datetime: str = ""
+    # Invocation bookends captured by mlpstorage_py/_invocation.py
+    # (metadata.invocation_start_time / invocation_end_time). These exclude
+    # read-side framework startup and write-side post-benchmark cluster
+    # collection, and are the authoritative origins for the Rules.md §4.7.1
+    # inter-phase cache-flush gap (storage#714 / #782). Empty when a
+    # pre-bookend results dir is read; consumers fall back to
+    # run_datetime / end_datetime (the DLIO summary start / end).
+    invocation_start_time: str = ""
+    invocation_end_time: str = ""
 
 
 @dataclass
@@ -828,6 +837,7 @@ class DLIOResultParser:
             ci_data = metadata.get('cluster_information')
             if ci_data:
                 system_info = ClusterInformation.from_dict(ci_data, self.logger)
+        metadata = metadata or {}
         return BenchmarkRunData(
             benchmark_type=benchmark_type,
             model=model,
@@ -841,6 +851,8 @@ class DLIOResultParser:
             metrics=summary.get("metric"),
             result_dir=result_dir,
             accelerator=accelerator,
+            invocation_start_time=metadata.get("invocation_start_time", ""),
+            invocation_end_time=metadata.get("invocation_end_time", ""),
         )
 
     def _load_summary(self, result_dir: str) -> Optional[Dict]:
@@ -962,6 +974,8 @@ class ResultFilesExtractor:
             metrics=run_metrics,
             result_dir=result_dir,
             accelerator=metadata.get('accelerator'),
+            invocation_start_time=metadata.get('invocation_start_time', ''),
+            invocation_end_time=metadata.get('invocation_end_time', ''),
         )
 
     def _load_summary(self, result_dir: str) -> Optional[Dict]:
@@ -1064,6 +1078,14 @@ class BenchmarkRun:
     @property
     def end_datetime(self):
         return self._data.end_datetime
+
+    @property
+    def invocation_start_time(self):
+        return self._data.invocation_start_time
+
+    @property
+    def invocation_end_time(self):
+        return self._data.invocation_end_time
 
     @property
     def num_processes(self):
--- a/mlpstorage_py/rules/models.py
+++ b/mlpstorage_py/rules/models.py
@@ -25,6 +25,27 @@ if TYPE_CHECKING:
     from mlpstorage_py.rules.issues import Issue
 
 
+def _compute_final_collection_time(metadata: Optional[Dict], summary_end: str) -> str:
+    """Return the write-side node-release proxy for the §4.7.1 gap.
+
+    The latest post-benchmark cluster ``collection_timestamp`` in
+    ``metadata`` (at or after ``summary_end``) marks when the write nodes are
+    released, and stands in for ``invocation_end_time`` on results dirs that
+    predate that bookend (pre-#787). Reuses the same helper the standalone
+    submission checker uses so the reportgen path selects an identical
+    write-side origin. Returns ``""`` when unavailable.
+
+    Imported lazily to avoid a top-level circular import between
+    ``rules/models.py`` and ``submission_checker``.
+    """
+    if not metadata or not summary_end:
+        return ""
+    from mlpstorage_py.submission_checker.checks.helpers import (
+        _latest_final_collection_timestamp,
+    )
+    return _latest_final_collection_timestamp(metadata, summary_end) or ""
+
+
 @dataclass(frozen=True)
 class RunID:
     """Identifier for a benchmark run."""
@@ -93,6 +114,14 @@ class BenchmarkRunData:
     # run_datetime / end_datetime (the DLIO summary start / end).
     invocation_start_time: str = ""
     invocation_end_time: str = ""
+    # Write-side node-release proxy for pre-#787 results dirs that lack
+    # invocation_end_time: the latest post-benchmark cluster
+    # collection_timestamp in metadata.json (see
+    # submission_checker/checks/helpers._latest_final_collection_timestamp).
+    # This is the middle §4.7.1 write-origin tier between invocation_end_time
+    # and the summary end; empty when no post-benchmark collection timestamp
+    # is recorded.
+    final_collection_time: str = ""
 
 
 @dataclass
@@ -838,12 +867,13 @@ class DLIOResultParser:
             if ci_data:
                 system_info = ClusterInformation.from_dict(ci_data, self.logger)
         metadata = metadata or {}
+        summary_end = summary.get("end", "")
         return BenchmarkRunData(
             benchmark_type=benchmark_type,
             model=model,
             command=command,
             run_datetime=summary.get("start", ""),
-            end_datetime=summary.get("end", ""),
+            end_datetime=summary_end,
             num_processes=summary.get("num_accelerators", 0),
             parameters=hydra_workload_config,
             override_parameters=override_parameters,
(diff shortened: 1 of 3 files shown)


wolfgang-desalvador:
This is again, down to the report, the same issue with the gap calculation @FileSystemGuy . Reported it here #825

FileSystemGuy:
Thanks for this — the mechanism is right, and the model-plumbing of the invocation bookends is exactly what the reportgen path was missing. I'd like to accept this approach, but there's one gap I think we need to close before it fully fixes #825 for the v3.0 round.

## The constraint: older submissions are eligible, and they have no bookends

For v3.0 we are accepting previously-generated results with **no regeneration required**. That matters here because the invocation bookends are very recent additions to the v3.0 codebase, not a baseline every result dir carries:

- `invocation_start_time` (read side) landed in **#714** (`b2dc7d9`)
- `invocation_end_time` (write side) landed in **#787 / storage#782** (`7b44f61`)

So a large fraction of eligible submissions were generated *before* #787 and have `invocation_end_time == ""` (and some, pre-#714, also have `invocation_start_time == ""`).

## The flaw this creates in the current PR

For a pre-#787 result dir, the write-side fallback chain collapses straight to the summary end:

```python
write_end = (
    _parse_summary_timestamp(ordered[0].invocation_end_time)  # "" -> None
    or _parse_summary_timestamp(ordered[0].end_datetime)      # -> summary "end"
)
```

That is exactly the measurement #825 is about — it charges the write-side post-benchmark cluster collection against the 30s cache-flush budget. And because that collection window scales with node count, the large-topology submissions are precisely the ones that will still trip the false `4.7.1` violation. So as written, the PR fixes #825 for freshly-generated (bookend-carrying) results but silently leaves it open for the older, larger eligible submissions.

The standalone submission checker already handles those older dirs — it has a **middle fallback tier** between the bookend and the summary end:

- `submission_checker/checks/checkpointing_checks.py` (the `4.7.1 checkpointCacheFlushValidation` check) selects the write-side origin as: `invocation_end_time` → **latest post-benchmark `collection_timestamp`** (`_latest_final_collection_timestamp(...)`) → `summary end`.

The reportgen path in this PR only has tiers 1 and 3, so it diverges from the standalone checker for exactly the pre-#787 vintage. (That middle tier exists specifically *because* pre-bookend dirs are expected to validate — which is the same situation we're in for this round.)

## Suggested enhancement

Add the middle tier to the reportgen path so it matches the standalone checker for the write side:

1. **Carry a precomputed write-side proxy on the model.** Add a field to `BenchmarkRunData` (e.g. `final_collection_time: str = ""`) and populate it in both `parse()` and `_from_metadata()` — both already receive the full `metadata.json` dict — by reusing the existing helper:

   ```python
   from mlpstorage_py.submission_checker.checks.helpers import _latest_final_collection_timestamp
   # summary_end is the same value used for end_datetime
   final_collection_time = _latest_final_collection_timestamp(metadata, summary_end) or ""
   ```

   `rules/models.py` already imports from `submission_checker` (`dlio_summary_helpers`, line ~603), so this follows the existing dependency direction — no new coupling.

2. **Insert it as the middle write-side tier** in the checker:

   ```python
   write_end = (
       _parse_summary_timestamp(ordered[0].invocation_end_time)      # #787+ dirs
       or _parse_summary_timestamp(ordered[0].final_collection_time) # pre-#787: node-release proxy
       or _parse_summary_timestamp(ordered[0].end_datetime)          # oldest: summary end
   )
   ```

   The formats are compatible: `_latest_final_collection_timestamp` returns a naive ISO string, and `_parse_summary_timestamp` tries `datetime.fromisoformat(value)` first, so it parses cleanly.

3. **Read side can stay as you have it** (`invocation_start_time` → `run_datetime`). There's no equivalent proxy for read-side framework startup, and its overhead doesn't scale with node count, so it's far less likely to push a real run over 30s. One nuance to be aware of (not necessarily to solve in this PR): the standalone checker *downgrades* a breach to a warning when it had to fall back to the read summary-start origin, whereas the reportgen path emits a hard INVALID. If you'd rather not diverge there, one option is to not emit an INVALID from reportgen when the authoritative read origin is absent — but that's a separate, lower-priority decision.

4. **Regression test.** A `BenchmarkVerifier`-level test over a pre-#787 fixture (bookend absent, per-node `collection_timestamp`s present) whose summary-gap > 30s but proxy-gap ≤ 30s, asserting the result is now `CLOSED` rather than `INVALID`. The `conftest.py` checkpointing fixtures (`chkpt_write_teardown_seconds`, `chkpt_omit_invocation_start_time`) already produce this shape.

With the middle tier added, this PR would fix #825 for the whole eligible set — including the large-topology older submissions that are the ones most likely to hit it — while staying a bounded change (one field + one helper reuse + one fallback line on top of what you have). Happy to pair on it or push a follow-up commit if that's easier. Thanks again!

FileSystemGuy:
@wolfgang-desalvador Just in case you weren't notified by github that I added a response to your PR...

wolfgang-desalvador:
> Thanks for this — the mechanism is right, and the model-plumbing of the invocation bookends is exactly what the reportgen path was missing. I'd like to accept this approach, but there's one gap I think we need to close before it fully fixes #825 for the v3.0 round.
> 
> ## The constraint: older submissions are eligible, and they have no bookends
> For v3.0 we are accepting previously-generated results with **no regeneration required**. That matters here because the invocation bookends are very recent additions to the v3.0 codebase, not a baseline every result dir carries:
> 
> * `invocation_start_time` (read side) landed in **[fix(#696): §4.7.1 gap origin = read invocation start, not summary start #714](https://github.com/mlcommons/storage/pull/714)** (`b2dc7d9`)
> * `invocation_end_time` (write side) landed in **[fix(storage#782): exclude write-side teardown from §4.7.1 failover-callout gap #787](https://github.com/mlcommons/storage/pull/787) / storage#782** (`7b44f61`)
> 
> So a large fraction of eligible submissions were generated _before_ #787 and have `invocation_end_time == ""` (and some, pre-#714, also have `invocation_start_time == ""`).
> 
> ## The flaw this creates in the current PR
> For a pre-#787 result dir, the write-side fallback chain collapses straight to the summary end:
> 
> ```python
> write_end = (
>     _parse_summary_timestamp(ordered[0].invocation_end_time)  # "" -> None
>     or _parse_summary_timestamp(ordered[0].end_datetime)      # -> summary "end"
> )
> ```
> 
> That is exactly the measurement #825 is about — it charges the write-side post-benchmark cluster collection against the 30s cache-flush budget. And because that collection window scales with node count, the large-topology submissions are precisely the ones that will still trip the false `4.7.1` violation. So as written, the PR fixes #825 for freshly-generated (bookend-carrying) results but silently leaves it open for the older, larger eligible submissions.
> 
> The standalone submission checker already handles those older dirs — it has a **middle fallback tier** between the bookend and the summary end:
> 
> * `submission_checker/checks/checkpointing_checks.py` (the `4.7.1 checkpointCacheFlushValidation` check) selects the write-side origin as: `invocation_end_time` → **latest post-benchmark `collection_timestamp`** (`_latest_final_collection_timestamp(...)`) → `summary end`.
> 
> The reportgen path in this PR only has tiers 1 and 3, so it diverges from the standalone checker for exactly the pre-#787 vintage. (That middle tier exists specifically _because_ pre-bookend dirs are expected to validate — which is the same situation we're in for this round.)
> 
> ## Suggested enhancement
> Add the middle tier to the reportgen path so it matches the standalone checker for the write side:
> 
> 1. **Carry a precomputed write-side proxy on the model.** Add a field to `BenchmarkRunData` (e.g. `final_collection_time: str = ""`) and populate it in both `parse()` and `_from_metadata()` — both already receive the full `metadata.json` dict — by reusing the existing helper:
>    ```python
>    from mlpstorage_py.submission_checker.checks.helpers import _latest_final_collection_timestamp
>    # summary_end is the same value used for end_datetime
>    final_collection_time = _latest_final_collection_timestamp(metadata, summary_end) or ""
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    `rules/models.py` already imports from `submission_checker` (`dlio_summary_helpers`, line ~603), so this follows the existing dependency direction — no new coupling.
> 2. **Insert it as the middle write-side tier** in the checker:
>    ```python
>    write_end = (
>        _parse_summary_timestamp(ordered[0].invocation_end_time)      # #787+ dirs
>        or _parse_summary_timestamp(ordered[0].final_collection_time) # pre-#787: node-release proxy
>        or _parse_summary_timestamp(ordered[0].end_datetime)          # oldest: summary end
>    )
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    The formats are compatible: `_latest_final_collection_timestamp` returns a naive ISO string, and `_parse_summary_timestamp` tries `datetime.fromisoformat(value)` first, so it parses cleanly.
> 3. **Read side can stay as you have it** (`invocation_start_time` → `run_datetime`). There's no equivalent proxy for read-side framework startup, and its overhead doesn't scale with node count, so it's far less likely to push a real run over 30s. One nuance to be aware of (not necessarily to solve in this PR): the standalone checker _downgrades_ a breach to a warning when it had to fall back to the read summary-start origin, whereas the reportgen path emits a hard INVALID. If you'd rather not diverge there, one option is to not emit an INVALID from reportgen when the authoritative read origin is absent — but that's a separate, lower-priority decision.
> 4. **Regression test.** A `BenchmarkVerifier`-level test over a pre-[fix(storage#782): exclude write-side teardown from §4.7.1 failover-callout gap #787](https://github.com/mlcommons/storage/pull/787) fixture (bookend absent, per-node `collection_timestamp`s present) whose summary-gap > 30s but proxy-gap ≤ 30s, asserting the result is now `CLOSED` rather than `INVALID`. The `conftest.py` checkpointing fixtures (`chkpt_write_teardown_seconds`, `chkpt_omit_invocation_start_time`) already produce this shape.
> 
> With the middle tier added, this PR would fix #825 for the whole eligible set — including the large-topology older submissions that are the ones most likely to hit it — while staying a bounded change (one field + one helper reuse + one fallback line on top of what you have). Happy to pair on it or push a follow-up commit if that's easier. Thanks again!

Working on it @FileSystemGuy

FileSystemGuy (approved):

PR #828, opened 2026-07-21
3rd: fix(reportgen/validate): four warnings-only review-tooling fixes (C1/C2/P5/P6)

FileSystemGuy:
## Summary

Third branch in the review-tooling stack. Four low-blast-radius correctness fixes to `reportgen` / `validate`, applied so they can be used **during** the v3.0 submission-review window while submitters stay on v3.0-rc1. Every change is **warnings-at-most** — none introduces a new hard-fail and none can flip a submission `pass → fail`.

**Stacked on top of:**
- #826 — reportgen final-results tables (`reportgen-v3-final-results-tables`)
- #827 — validate flaw warnings (`validate-flaw-warnings`) ← **base of this PR**

Review this PR's diff against `validate-flaw-warnings`; it isolates just the commits below.

## Changes

| Item | Fix |
|------|-----|
| **C1** — `JSONParser.__contains__` | Referenced `self.messages` (nonexistent) → `AttributeError`. Now uses `self.d`. Was latent because callers reach `[]`/`.get`, never `in`. |
| **C2** — 4.3.4 aggregate accelerator memory | An aggregate-memory shortfall hard-failed the submission. The 80 GiB H100 per-accelerator figure is only a baseline (checkpointing metadata carries no accelerator-type field), so it now surfaces via `warn_violation` (with the accelerator count + a pointer to verify against the real accelerator memory) and returns `True`. |
| **P6** — metric-block filtering | `DLIOResultParser.parse` stored `summary['metric']` unfiltered, so a scalar/string entry could reach `report_generator`'s `fmean(metric_list)` and raise `TypeError`. New `_filter_list_metrics()` filters to list-valued keys on **both** parse paths and logs a diagnostic when keys are dropped or the block collapses to nothing — a blank metrics column is now traceable to a logged reason instead of vanishing silently. |
| **P5** — kvcache total data loss | `_aggregate_option_results` folded an empty trial's `sum([]) == 0.0` into the across-trial mean and coerced a fully-empty aggregate to `0.0`, making a **total data loss** indistinguishable from a real measured zero. It now skips no-data trials and emits `None` (blank) on total loss. A present-rank `0.0` (CPU-tier working set) is a legitimate measured value and is still counted. |

## Testing

TDD RED→GREEN for each item. Full CI-mirror sweep green after every commit:

- `tests/` — 2970 passed, 1 skipped
- `mlpstorage_py/tests` — 898 passed
- `vdb_benchmark/tests` — 228 passed
[...]

PR #829, opened 2026-07-21
4th: feat(reportgen): fixed webpage-parity schema for results.{csv,json}

FileSystemGuy:
## 4th (stacked on #828): fixed webpage-parity schema for `results.{csv,json}`

Stacked on top of `reportgen-validator-fixes` (**#828**). Part of the reportgen review-tooling stack #826 ← #827 ← #828 ← **this**. Retarget base to `main` once the lower PRs land.

### What

`reports reportgen` now emits `results.csv` / `results.json` as a **fixed contract** — exactly the reference columns for all 8 v3.0 results-web-page tables (`Results Table Structure.xlsx`) plus three discriminator columns (`Division`, `Benchmark Type`, `Model`) so one flat file can carry every table's rows. The column set is **never data-driven**: a MLCommons staff member opens `results.csv` in Excel and reduces it to any single web table by deleting the workload blocks + discriminator columns that don't apply.

### How

**Output projection, not an aggregation rewrite.** The internal `_aggregate_*` helpers (`train_mean_of_*` / `checkpoint_mean_of_*` / `vdb_*` / `kvcache_*` machine keys, and the empty-metric `StatisticsError` → INVALID gate) are **untouched**, so validity behavior is unchanged. A new `_final_row()` runs inside `write_json_file` / `write_csv_file` and cherry-picks each fixed column from the in-memory row. The dynamic `*_mean_of_*` per-metric means and the trailing `issues` column are simply no longer emitted — this is an **output-only** change with no path from pass to fail.

### Schema (54 columns)

- **Left edge (13):** `Public ID`, `Organization`, `Division`, `Benchmark Type`, `Model`, `Name`, `Description`, `Type`, `Access Protocol`, `Availability`, `RU's`, `Integrated Client Storage`, `Usable Capacity (TiB)`.
- **Workload blocks:** `Training - …`, `Checkpointing - …`, `VDB - …`, `KVCache - …`. Shared reference names (`# Client Nodes`, `Code`, `Read B/W (GiB/s)`, …) are workload-qualified so every column key is unique.
- `Division` = category upper-cased (`CLOSED`/`OPEN`); `Model` = reference display label (`llama3-1t` → `1250B`), blank for kvcache/vdb (single-table workloads).
- `Name` shows the system name and links to `system-description.yaml`; `Description` shows `PDF` and links to the `.pdf`.
- `Code`/`Logs` are **per-workload** (an OPEN row may ship its own code), routed into the row's own block from the run's code-image pointer.
- KVCache option `1`/`2`/`3` map to the reference groups `llama3.1-8b Storage Only` / `Storage + Mem` / `llama3.1-70b Storage Only`.
- Removed: the machine-key prefix (`category`/`orgname`/`systemname`/`benchmark_type`/`model`/`accelerator`), the `issues` column, and the now-dead `_ordered_fieldnames` helper.

### Tests

RED-first. Output-file tests migrated to the fixed-schema contract; internal-structure tests (aggregation helper keys) unchanged. All 4 CI suites green: `tests/` 2975p·1skip, `mlpstorage_py/tests` 898p, `vdb_benchmark/tests` 228p, `kv_cache_benchmark/tests` 238p.


FileSystemGuy:
**R5 (pushed `79b9833` + `eceb601`): KVCache "P95 Read Latency (ms)" was surfacing the wrong source metric.**

The column mapped `summary.storage_io_latency_ms.p95`, which the kv_cache benchmark computes as the P95 of **per-request cumulative** storage I/O time — the initial cache read, plus one read per decode batch (potentially hundreds per request), plus write latency on misses, all summed into a single number per request (`kv_cache/benchmark.py:596-632, 1438`). For Storage Only runs this inflated the column ~10–500× (e.g. TTA showed 461164 ms), and a high-cache-hit run legitimately produced 0.0 (hpe: >95% of requests did no storage I/O at all).

The genuine per-IO read percentile already exists in every rank result file as `cache_stats.storage_read_p95_ms` (`kv_cache/cache.py:1013-1017`) and was previously unread.

Fix (TDD RED `79b9833` → GREEN `eceb601`):
- `_aggregate_option_results` now emits `aggregated_storage_read_p95_ms` (max across ranks/trials, matching the existing convention; ranks that never touched storage are skipped; `None` when absent everywhere — never a fabricated 0.0).
- The fixed-schema column now sources that field.
- For summary.json files that predate the field (every existing v3.0 submission), reportgen backfills it from the per-rank `kvcache_results_*.json` files that ship inside each submission — **no submitter rerun required**. A present summary field always wins over recomputation.

No schema change: all emitted `results.{csv,json}` (global, per-org, per-workload) remain the same 54 columns; only the value in the existing column changed. Verified against the real v3.0 tree: TTA Storage Only 461164.28 → 90460.08 ms; hpe Storage Only 0.0 → 11787.55 ms. All 4 CI suites green.

FileSystemGuy:
All commits in this PR are now on main: the reportgen-column-parity integration branch (which contained this PR's head) was fast-forward merged into main at b15308a, tagged v3.0-review-final. GitHub cannot mark a stacked PR merged when its head is already an ancestor of main, hence closed-not-merged — the code itself is fully merged. See #826 (auto-marked merged) for the stack's mainline record.

PR #837, opened 2026-07-30
fix(reportgen): warn when a kv_cache row hides a collapsed group or a failed run

### LINKED ISSUE #836: reportgen: non-canonical kv_cache run layout publishes one run's metrics and silently discards the rest
## Summary

When a kv_cache workload directory contains multiple run directories that do **not** match the canonical `<model>/run/<timestamp>/` layout, `reportgen` publishes a results row built from **one** of those runs and silently discards the rest. The only diagnostic emitted is `No valid run directories found`, which describes neither the row that was produced nor the measurements that were dropped.

This is worse than dropping the workload entirely: a blank row is visible to a reviewer, whereas a populated row assembled from one of several measured configurations is not.

## Observed in

`mlcommons/submissions_storage_v3.0` @ `316717f2e`, ANL `crux-eagle`, tool version 3.0.46.

The submission organizes its kv_cache runs by cluster topology instead of by run timestamp:

```
closed/ANL/results/crux-eagle/kv_cache/llama3-8b-10u/
├── 1nodex8ppn/    (8 ranks)    summary.json + kv_cache_20260723_062638_metadata.json
├── 8nodex8ppn/    (64 ranks)   summary.json + kv_cache_20260723_191548_metadata.json
└── 64nodex8ppn/   (512 ranks)  summary.json + kv_cache_20260724_051511_metadata.json
```

All three directories are complete: each has a `summary.json` with options 1/2/3, option subdirectories, a `*_metadata.json`, and 389 per-rank result files between them. Nothing is missing from the submission.

## Behavior

All three runs are discovered and verified. From the reportgen log:

```
INFO: Created benchmark run: kv_cache_run_llama3.1-8b_20260723_062638
INFO: Created benchmark run: kv_cache_run_llama3.1-8b_20260724_051511
INFO: Created benchmark run: kv_cache_run_llama3.1-8b_20260723_191548
STATUS: Benchmark run qualifies for CLOSED category (... run_datetime='20260723_062638')
STATUS: Benchmark run qualifies for CLOSED category (... run_datetime='20260724_051511')
STATUS: Benchmark run qualifies for CLOSED category (... run_datetime='20260723_191548')
WARNING: No valid run directories found in .../crux-eagle/kv_cache/llama3-8b-10u
```

The resulting tables disagree with each other:

| Table | Rows |
|---|---|
| `crux-eagle/kv_cache/llama3-8b-10u/results.csv` | **0** (and `results.json` is `[]`) |
| `crux-eagle/kv_cache/results.csv` | 1 |
| top-level `results.csv` | 1 (`v3.0-0005`) |

The workload-level table is empty, consistent with the warning. But the system rollup and the global table both carry a row — and its values are `1nodex8ppn`'s alone:

| Cell | Published in `v3.0-0005` | `1nodex8ppn` | `8nodex8ppn` | `64nodex8ppn` |
|---|---|---|---|---|
| 8b Storage Only tok/s | 5865.99 | **5865.99** | 48.57 | 0.0 |
| 8b Storage + Mem tok/s | 49.84 | **49.84** | 170.16 | 163.96 |
| 70b Storage Only tok/s | 2521.63 | **2521.63** | 1110.03 | 738.48 |
| 8b Storage Only Read B/W | 0.0 | **0.0** | 0.776 | 0.0 |
| `# Client Nodes` | 1 | **1** | 8 | 64 |

The 8-node and 64-node measurements — which have materially different bandwidth and latency figures — do not appear anywhere in any published table. The three zero-valued bandwidth cells and three blank P95 cells are likewise the 1-node run's own values, not placeholders.

Empirically the earliest-timestamped run wins, but I did not pin the selection rule.

## Apparent root cause

Two different paths disagree about what constitutes a run:

- **Row placement** uses the canonical layout. `_validate_model_dir` (`mlpstorage_py/reporting/directory_validator.py:411`) accepts a child of the model directory only if `_is_datetime_dir(entry.name)` or `entry.name in EXPECTED_COMMANDS`. `1nodex8ppn` is neither, so nothing is placed into the workload-level table and the warning at line 428 fires.
- **Rollup aggregation** consumes the extracted `BenchmarkRun` objects directly. Those were built successfully — each topology directory has a `*_metadata.json`, so `ResultFilesExtractor.extract` (`mlpstorage_py/rules/models.py:976`) identifies all three as kvcache runs without needing the directory name.

So the runs exist in memory, are excluded from the leaf table, and then collapse into a single row upstream because all three share the same `(program, command, model)` grouping key and differ only in `run_datetime`.

## Why the warning is inadequate

`No valid run directories found` is emitted against the workload directory, is a `WARNING` among 183 others, and names no consequence. Nothing in the output states that a row was published, that it represents one of three configurations, or which two were dropped. In a 179-row table this is not something a reviewer can be expected to catch — the row looks complete.

## Suggested fixes

Independent of any decision about whether the layout itself should be accepted:

1. **Never silently reduce N measured configurations to one.** If multiple runs group to one row and their metrics disagree, that is either an error or a row per configuration — not a silent pick.
2. **Make the warning name the consequence**, in the style of #835: say that a row was published from a subset, how many runs were dropped, and which.
3. **Make the two paths agree.** A run good enough to extract, verify and aggregate into a rollup should not be invisible to the leaf table that is supposed to contain it — the disagreement between an empty `llama3-8b-10u/results.csv` and a populated parent rollup is itself a bug.
4. **Submission-checker gate**: a workload whose leaf table has zero rows while its rollup has one should fail validation.

Item 1 is the one that affects published results.

## Related

- #835 — silent drop of kv_cache/VDB run leaves missing `*_metadata.json`. Same family of problem (a run disappears without the output saying so), different trigger. This one is more severe because it produces a wrong number rather than a blank.
- The layout question itself is a separate WG matter, written up as open question 3 in `ApparentProblems.md`.

FileSystemGuy:
Fixes #836.

A kv_cache row could be built from one of several runs in its workload group, chosen by filesystem-discovery order, with no indication on the row. A second, unrelated path let a run that lost most of its per-rank result files publish a row shaped exactly like a clean one.

Base is `reportgen-column-parity` (the integration branch) rather than `main` — these touch the same reportgen files the #826–#829 stack rewrote.

## What changed

**Deterministic selection.** `_select_kvcache_run` replaces `runs[0]`. Discovery order was genuinely arbitrary: of the three groups in the v3.0 tree holding two or more real runs, it published the earlier run for two and the later run for the third. Earliest `run_datetime` has no better claim than latest, but it is a rule rather than an accident, and it matches the positional convention Rules.md §2.1.17 already uses for training.

**`datasize` excluded from selection.** The kv_cache grouping key has no `command` component (D-05), so a `datasize` leaf lands in the same group as the `run` it sized. It writes no `summary.json`. An earlier draft of this change selected those leaves and blanked all twelve KVCache metrics on rows v3.0-0019 / 0021 / 0023 / 0142; the expectation diff caught it and it is now pinned by a test.

**Two warnings on `Result.issues`.** One when a row is built from a subset of its group, naming the published run and every dropped one; one when the published run recorded `partial_failure`, with the missing-file and failed-trial counts. The collapsed-group warning requires two or more actual `run` invocations, so ANL polaris-nvme's pair of `datasize` leaves stays quiet.

**`format_issues_list(show_all=False)` now shows CLOSED warnings.** The filter dropped every issue whose `validation` was `CLOSED`, conflating two orthogonal fields — `validation` is the submission category, `severity` is whether a human needs to look. Without this the new warnings reach only the log stream, since the fixed webpage-parity schema has no issues column. The datagen leaf-presence warnings, written to be "worth surfacing but not invalidating", had been invisible for the same reason and now print. Warnings are badged `[WARN]` in yellow rather than green `[CLOSED]`, which on "row built from 1 of 3 runs" reads as approval; the stray `None: ` placeholder is dropped for issues carrying no `parameter`.

## Findings behind the change

The issue's original diagnosis was wrong on two counts; the correction is [posted there](https://github.com/mlcommons/storage/issues/836#issuecomment-5124505505) in full.

`crux-eagle` is not a novel layout the tool failed to parse. Each run's `metadata.result_dir` records the path the tool wrote it to — `crux-eagle`, `crux-eagle-n8`, `crux-eagle-n64`, each canonical `<system>/kv_cache/llama3.1-8b/run/<ts>/`. Three system directories were merged into one at packaging time. Accepting the layout would not fix the collapse, because `systemname` is path-derived and all three still produce the same grouping key.

The two dropped runs are failed runs, not lost measurements: `8nodex8ppn` is missing 403 per-rank result files (~70 %) and `64nodex8ppn` 4473 (~97 %), both with 9 of 9 mpirun trials exiting non-zero. `partial_failure` and `trial_failures` are written by `kvcache.py:919,960` and were read by nothing, which is the more general bug — `open/farmgpu/.../llama3.1-70b-instruct/run/20260709_215001` publishes today missing 45 rank files.

## Verification

Five commits, TDD RED→GREEN throughout. All four CI suites green: 3080 + 911 + 238 + 228 passed, 1 skipped, 0 failures.

Against the v3.0 tree the change produces 3 collapsed-group warnings (ANL crux-eagle 3 runs, Everpure 51hosts_20 and 51hosts_30 2 runs each) and 1 partial_failure warning, adding five lines to the printed summary — the four above plus one datagen leaf warning that had never been visible.

One published row moves: **v3.0-0019** (Everpure 51hosts_20) switches from its later run to its earlier one, 85736 → 66113 tok/s and 11 other cells, because discovery order had picked the later run there. Which run represents a submitter's intended submission is not something the tool can know; that question is being raised with the three affected submitters in `ApparentProblems.md` alongside the results refresh.

### DIFF
--- a/mlpstorage_py/report_generator.py
+++ b/mlpstorage_py/report_generator.py
@@ -52,6 +52,24 @@ _INVALID_MSG_EMPTY_METRIC = (
     "metric {key} is empty in invocation {basename}; cannot aggregate"
 )
 
+# Issue #836 kv_cache integrity warnings. Warnings, not INVALID gates — both
+# conditions describe a row that is misleading rather than one that violates a
+# Rules.md invariant, and the kv_cache branch has no rules-strict gates by
+# design (the D-22 pass-through boundary). They ride on ``Result.issues`` so
+# they reach the published ``issues`` column; a log line alone is not something
+# a reviewer of a 179-row table can be expected to catch.
+_WARN_MSG_KVCACHE_COLLAPSED_GROUP = (
+    "[WARN] kv_cache row built from 1 of {n} runs in this workload group: "
+    "published {published}; dropped {dropped}. The dropped runs' metrics "
+    "appear in no published table."
+)
+_WARN_MSG_KVCACHE_PARTIAL_FAILURE = (
+    "[WARN] kv_cache run {run} recorded partial_failure in summary.json: "
+    "{missing} missing per-rank result file(s), {failed} mpirun trial(s) with "
+    "non-zero exit. Published metrics are aggregated over the surviving "
+    "ranks only."
+)
+
 
 class Hyperlink:
     """A results-table cell that is a (visible-text, href) pair.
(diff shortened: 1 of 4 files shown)


FileSystemGuy:
Closing — these commits belong directly on `reportgen-column-parity`, which is serving as the working main branch until the submission round completes. Pushed there instead.

